PR #1, opened 2020-05-26
Re-implement CI with GitHub actions

jrozner:


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.13
```
JoelAtDeluxe, line 19: Internally, the go code is set at 1.12. I was actually wondering if we should update. Thoughts?

jrozner: Ya, I think it makes sense. I've been going through the actions documentation and this is just a template to start with. Working on getting everything migrated over and can take care of that as well. In terms of Go support they're pretty good about maintaining backwards compatibility but I'm find with testing and officially supporting whatever the current stable is.

PR #3, opened 2020-06-04
Add archive-viewer frontend build target

alexdavid:
Based on #2 

This PR adds a frontend build target for the local archive viewer to `backend/archive_templates` where @JoelAtDeluxe's work will pick it up and bundle it up in the zip for archival.

PR #6, opened 2020-06-08
Allow backend to export operations to file

JoelAtDeluxe:
This PR provides a mechanism to export an operation to a zip file. The file is placed in a new archive content store, and can be reviewed by retrieving the export from the archive content store,  and then opening the archive viewer on a user's computer. This is a continuation of PRs #2  and #3. 

## UI Changes

* A new option to export an operation has been added to the administrator's operation list.
* Clicking this button queues an export with the database. 
* Only one export can be enqueued at a time. This is enforced on the frontend by disabling the archive button for any pending archives. (the backend has it's own mechanism for enforcing a single enqueued export)
* Once an export has been started, it can be re-archived, and the frontend button will be re-enabled. Note: If a user simply waits on this page, the button will not re-enable. A page refresh is required. 

## Backend Changes

* A new "Archive" data store has been introduced. This can be the same data store that is used for evidence, however, it has been separated to allow for easier review of archives (so that they are not interspersed with random evidence).
* A new export processor has been added, and enabled on the dev web server. The processor checks the database exports queue, and once at least one item is in the queue, it will, sequentially, start generating exports, writing them (mostly) directly to the content store.
* Content stores have been updated to allow uploading a file with a specific name, to enable exports
* The backend enforces a single enqueued export by checking for the existence of a record (matching operation id and operation status). Should it find a match, the insert will fail. Otherwise, the insert succeeds, ensuring we always have 0 or 1 enqueued records for the (operation_id, "pending" status id ) tuple

## Notes

* The exporter should allow for concurrent exports over multiple processes. The processor will perform row-locking to ensure that only one processor can claim an export at a time. In the dev webserver, only a single processor exists.
* For production purposes, there is currently no way to export. We need to decide if one instance of Web equates to one instance of the exporter (as is done on the dev server), or if the exporter needs another main file, as with the api server. 
* While most files are streamed from the normal/evidence content store directly to the archive content store, terminal recordings, due to their format, need to be handled specially. These will be read completely from the evidence store, then sanitized (converted into a single, long json string), then written out. As such, the memory requires for the processor grow to include the size of the biggest terminal recording, plus a bit more.
* This solution is mysql specific, and requires innodb to provide the row-level locks.

### DIFF
--- a/backend/processors/export_operation_internal_test.go
+++ b/backend/processors/export_operation_internal_test.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"io/ioutil"
 	"testing"
 
 	"github.com/jonboulle/clockwork"
@@ -66,6 +67,20 @@ func TestCopyStreamToZip(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, filename, creator.StreamName)
 	require.Equal(t, content, creator.Stream.String())
+}
+
+func TestCopyFileToZip(t *testing.T) {
+	creator := newSimpleCreator()
+	src := "export_processor.go"
+	dst := "result_file.go"
+
+	data, err := ioutil.ReadFile(src)
+	require.NoError(t, err)
+
+	err = copyFileToZip(&creator, src, dst)
+	require.NoError(t, err)
+	require.Equal(t, dst, creator.StreamName)
+	require.Equal(t, data, []byte(creator.Stream.String()))
 }
 
 func TestCopyContentStoreFile(t *testing.T) {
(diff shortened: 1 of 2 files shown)


alexdavid:
For debugging locally you can run the following to see a list of exported archives:
```
$ docker-compose exec backend sh -c 'ls /tmp/contentstore/*.zip'
```
And then "download" one:
```
$ docker cp $(docker-compose ps -q backend):/tmp/contentstore/{zip-file-name}.zip .
```
Alternatively you can check `/var/lib/docker/volumes` on linux, however I don't think this is exposed on macos

JoelAtDeluxe:
Additionally, you'll want to generate the necessary assets make the asset viewer work. You can do this via:
`cd frontend && npm run build-archive-viewer`

PR #9, opened 2020-06-18
Revert data source changes

jrozner:
This reverts a commit that was made before the project was squashed and
open sourced removing the concept of data sources due to changing
requirements in an effort to simply the code. A significant amount of
the work to support operation archive has already been completed and we
can keep it around in a branch, with a rever of this commit, as the base
in case we decide to pick up the feature again.


backend/dtos/dtos.go
```diff
@@ -51,10 +51,7 @@ type Operation struct {
 }
 
 type Query struct {
-	ID    int64  `json:"id"`
-	Name  string `json:"name"`
-	Query string `json:"query"`
-	Type  string `json:"type"`
+	ID int64 `json:"id"`
```
JoelAtDeluxe, line 54: I think this change is unrelated to data source change. This is simply returning more data, which it looks like the front end uses

JoelAtDeluxe: Yes, that's right. If you look at list_queries_for_operation, it defines this struct:
```go
type QueriesForOperationOutput struct {
	ID    int64  `db:"id" json:"id"`
	Name  string `db:"name" json:"name"`
	Query string `db:"query" json:"query"`
	Type  string `db:"type" json:"type"`
}
```
This is effectively the same data dtos.Query is providing. I think this change was made because we wanted all return types to live in the dtos package, rather than creating one off instances inside the individual services.

frontend/src/pages/operation_show/finding_modals/index.tsx
```diff
@@ -41,6 +41,7 @@ export const CreateFindingModal = (props: {
     onSuccess: () => props.onRequestClose(),
     handleSubmit: async () => {
       const finding = await createFinding({
+        occurredAt: new Date,
```
JoelAtDeluxe, line 44: this isn't used, which is probably why it was removed

jrozner:
Per discussion with @JoelAtDeluxe and @alexdavid we're going to keep this change in place. It provides the ability to mock services for the frontend to make testing business logic easier. An additional patch will be created against master that moves the dtos generation into webpack so that we can generate on each build rather than relying on what is checked in to be the correct version.

Issue #8, opened 2020-06-17
Move evidence between operations


jrozner:
Provide the ability to move evidence from one operation to another. This is not as straightforward as just changing the foreign key in the row because the associated tags are references to tags that exist in the operation and not globally. Based on the recently added work to allow tagging at creation time, we have two options:
1. Delete tags that exist then move the evidence over
2. Try to match referenced tags to existing ones in the other project or create new ones, if they don't exist, remove existing tag references, the move the evidence over.

I'm open to either currently.

JoelAtDeluxe:
Questions on the semantics here:
1. Is this sharing evidence, or moving, or either (meaning: are we implementing this because someone uploaded evidence to the wrong operation, or are we implementing this because two operations happen to need the same evidence... or both?)
2. (in case this is a proper move) What do we do if the evidence is paired with a finding? Do we disallow moves here, or do we remove this evidence from the finding? I'm not sure how viable trying to re-associate moved evidence with some pre-existing finding... unless that finding consists entirely of that single evidence.

Presumably all other data-oriented aspects are copied over, which should be pretty trivial.

JoelAtDeluxe:
With respect to mapping tags, I suppose we'd compare names, and if they were exact matches (maybe ignoring spacing and case) we'd re-associate the evidence with the "new" tag. I'm somewhat concerned that generating _actual_ new tags would create a kind of tag bloat, where you end up with two tags that are functionality identical, but technically separate. A guided process to re-map tags seems excessive to me, so I suppose I'd recommend we do the following with respect to re-tagging:

1. Get a list of associated tag names for the specific evidence
2. disassociate all tags from that evidence
3. Get a list of all tags for the "new" operation
4. For each tag, check if there's an approximate match.
   * If so: associate that tag with the moved evidence
   * If not: ignore that tag

This should help limit tag bloat, while allowing some reduced effort in migrating evidence.

JoelAtDeluxe:
Small recap from offline discussion:
1. For tags, we will try to match when tag names are close, and drop tags that are not close. Closeness metric is TBD, but likely something along the lines of mentioned above.
For tags that are not matched, we may opt to display what tags are being dropped, allowing users an easy to to re-add tags if they are missing.
2. For findings, we will drop evidence from the finding. It will be incumbent on the user migrating evidence to re-associate evidence with a finding in the new operation.

JoelAtDeluxe:
Addressed in PR #12

PR #12, opened 2020-06-23
Allow moving evidence between operations

JoelAtDeluxe:
# Move evidence between operations

This PR introduces the ability to migrate evidence from one operation to another, while attempting to keep as much tagging as possible. This is included as a solution for when users accidentally add evidence to the wrong operation.

Addresses Issue: #8 

### Who can use this

Users who have at least a write permission to both the source operation and the destination operation

### Back End Changes

* Adds two new endpoints: one to do a kind of pre-check of the migration, and one to do the actual moving.
* Adds two new types to describe the pre-check
* Everything else is in support of the frontend changes

### Front End Changes

* Restructures the timeline actions:
   * Not all actions can fit in the button group, so "extra actions" are now included the "Coy Permalink" button, and represented with a downward arrow. Clicking this presents a drop down menu with each "extra" action (similar to how the admin user table actions are used)
   * Added "move" as an extra action, and moved "Delete" into the extra actions
* Added a modal to preface the change, indicating what tags would be lost, as well as noting that this evidence will be dropped from all findings.
* After moving, user is redirected to the current page, refreshing the page.

## Meeting License Requirements

### DIFF
--- a/backend/dtos/gentypes/generate_typescript_types.go
+++ b/backend/dtos/gentypes/generate_typescript_types.go
@@ -30,6 +30,8 @@ func main() {
 	gen(dtos.UserOperationRole{})
 	gen(dtos.DetailedAuthenticationInfo{})
 	gen(dtos.SupportedAuthScheme{})
+	gen(dtos.TagPair{})
+	gen(dtos.TagDifference{})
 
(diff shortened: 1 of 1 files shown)


JoelAtDeluxe:
The above noted issue with this PR should now be fixed.

Issue #10, opened 2020-06-22
Add ability to remove operations


jrozner:
Provide the ability to remove existing operations

## Requirements
1. This action should remove all associated data with the operation: tags, findings, evidence, etc.
1. A new option should be added to the operation settings page to perform a deletion. This should require a confirmation (typing the operation name) in a modal and only be allowed by an operation admin. The modal should also warn that it is unsafe to redeploy a server during a delete operation

jrozner:
Implemented by #14

PR #14, opened 2020-06-24
Add ability to delete Operations

JoelAtDeluxe:
# Add operation-admins to delete their operations

Adds the ability for users to delete an operation and corresponding data. This will remove both database records as well as content store data.

Addresses Issue: #10 

### Who can use this

This is restricted to admins of the operation in question, as well as super admins, via the admin/operations list menu

### Back End Changes

* Adds a new service and endpoint to trigger the delete (this is a synchronous process, as it is expected that operations won't be sufficiently large to justify offloading this to a separate worker.)
* Updates the delete evidence endpoint to also delete content
* Works for all evidence types, including no-content evidence
* Adds some new test helpers

### Front End Changes

* Adds to the Operation settings a new Delete Operation button that kicks off the delete process
* Delete operation prompts the user with a simple challenge, as is done in the admin section (the requesting user will need to type in the operation slug to verify that they actually want to delete the operation)
* Returns the user to the list operations page upon successfully deleting an operation

## Meeting License Requirements

### DIFF
--- a/backend/services/seed_data_helpers_test.go
+++ b/backend/services/seed_data_helpers_test.go
@@ -33,7 +33,7 @@ var TinyImg []byte = []byte{
 	0x42, 0x60, 0x82,
 } // tiniest png https://github.com/mathiasbynens/small
 
-var TinyCodeblock []byte = []byte(`{"contentType": "codeblock", "contentSubtype": "python", "content": "print(\"Hello World!\")"}`)
+var TinyCodeblock []byte = []byte(`{"contentSubtype": "python", "content": "print(\"Hello World!\")"}`)
 
 var TinyTermRec []byte = []byte(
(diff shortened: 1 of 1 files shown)


JoelAtDeluxe:
Note that this solution does everything within the same request. This should be fine, as it's unlikely that remote deletes will take an excessive amount of time, however this could be revised to support huge operations as follows:

1. Upon starting delete, record all evidence and add this to a separate table in the database as a queue of keys to remove from the content store
2. Develop a worker that will read this queue and progressively remove items

The upside to this process is that delete operations should become much quicker, as it instead becomes about 10ish database queries, executed in a transaction. It also allows easy removal of orphaned files (though this really shouldn't be necessary). It should also allow us to remove the warning on redeployment, since it'll all be transaction based.
The downside is that another process handle this removal, and also that content would remain around longer. This could be done as a long running goroutine within the existing backend, or an entirely separate application that can be managed separately.

Issue #16, opened 2020-06-26
Admins do not 404 for operations they are not a member of


jrozner:
When a super admin force browses to an operation that they are not a member of they are able to load the page with a timeline though the media and left rail fails to load. From here they are able to interact with the operation and read content or force browse to the operation settings view where they can perform any action including adding themselves as a member with whatever permission is desired.

JoelAtDeluxe:
This is a side effect due to the way we implemented super admins being able to access an operation's settings page via the admin page. Essentially, we added in a check for permissions on a number of services. Roughly:

```python
if super_admin:
  allow
else:
  do_actual_permission_check()
```

I see 3 possible fixes right now (in reality, there could be more or fewer):

1. Remove the bypass above specifically for list evidence for operation. I'm not sure why this is there currently, and it may be required, but if not, we can achieve a kind of 404-lite experience (still able to view the page, but every sub-component renders a 404 error)
2. Add some kind of tag or indication on every request done via an admin that the admin bypass should be checked, which hopefully will prevent this situation, except when the admin specifically tries to do this.
3. Duplicate all of the endpoints to have admin versions and non-admin versions. 

There are potentially issues with each of these, so we may need to reconsider the whole feature. 

Two notes though: Currently, while super admins can view evidence (to a degree -- just tags and descriptions as far as I can tell), they cannot edit that evidence. Also admins should be able to view the settings page of any operation, per a previous request. How they get their is kind of up to them -- it's the same page regardless. Do we need two proper versions to prevent the above situation, knowing that super admins  have this ability anyway?

JoelAtDeluxe:
Regarding list evidence for operation:  This exists because this is the way we calculate how many pieces of evidence are tied to a particular tag: We get a list of all evidence, then do some typescript work to reduce that evidence into a count for each piece of evidence. I think we can revise this specific functionality to instead hit a different endpoint and have the backend do this work for us (either in Go, or hopefully in sql), and then removing the bypass on this service would be possible. This should actually make navigating to an existing, but non-accessible operation, look the same as a non-existing operation. 

However, as noted above, this would not affect settings, as the settings menu is specifically allowed to be accessed by super admins.

JoelAtDeluxe:
This is addressed in PR #28 

Note that this only affects manually browsing to the operation / evidence listing page. Super admins, by design, are still allowed to manually navigate to an operation settings page.

PR #28, opened 2020-07-15
Prevent super admins from looking at operations they do not belong to

JoelAtDeluxe:
This PR will generate a 404 response for super admins who try to manually navigate to operation for which they do not belong. 

Addresses Issue #16 

Background: 
Per the notes on Issue 16, the core problem is that listing evidence for an operation allows for an admin bypass. Super admins do not need to view evidence, they only need to view operation settings -- one facet of which is editing tags, and getting a usage count for each tag, which in tern got this data by listing all evidence for an operation. This PR removes that bypass for viewing evidence, and instead provides another mechanism to retrieve tag usage counts.

Details:
* Adds a new dto: TagWithUsage, which contains an `EvidenceCount` field which, appropriate, reflects the count of evidence using this tag
* Limits this dto response to just the listTagsForOperation. This is actually used in a handful of places, so some adjacent code has been introduced to convert this from a `dto.TagWithUsage` into a standard `dtos.Tag`
* Updates the frontend to know and use this new type, in addition to simplifying the logic to use the tag count when rendering the operation settings tag view.

### DIFF
--- a/backend/services/list_evidence_for_operation.go
+++ b/backend/services/list_evidence_for_operation.go
@@ -12,6 +12,7 @@ import (
 	"github.com/theparanoids/ashirt-server/backend/helpers"
 	"github.com/theparanoids/ashirt-server/backend/models"
 	"github.com/theparanoids/ashirt-server/backend/policy"
+	"github.com/theparanoids/ashirt-server/backend/server/middleware"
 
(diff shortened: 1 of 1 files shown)


jrozner:
Changes look good. Not sure whether this is more or less intuitive but I'm not as confused as I was when we first saw this.

Issue #39, opened 2020-08-06
Unable to select tags with cursor


jkennedyvz:
When editing a finding in the web ui a user can press the backspace key to remove the rightmost finding only. The cursor in the edit window does not allow selection of any preceding tags.

![image](https://user-images.githubusercontent.com/65985482/89592633-f11d0800-d801-11ea-9a01-b0cf2bea80dd.png)

JoelAtDeluxe:
I've created a solution, but I'm not sure it's actually desirable. Try it out and let me know what you think.

PR #40, opened 2020-08-07
Allow left/right tag navigation in tag chooser

JoelAtDeluxe:
This PR adds the ability to select a specific tag from the included tags in a tag chooser, and delete it via the backspace key (also now allows the delete key).

Addresses Issue #39 

You shouldn't need instructions, but here they are:

* While the input box is empty, pressing left or right arrow keys will select a tag in the list. 
   * When no tags are selected, pressing left will jump to the start of the list of tags, while pressing right will  jump to the end of the tag list.
* Pressing backspace or delete while a tag is selected will remove that tag. the cursor will be updated to move backwards towards the start of the list.
   * When the first tag is removed, no tags will be highlighted.
* Once a user types any other key besides left/right arrow, backspace, or delete, then the selection will removed. 
* If the input box is not empty, then left/right arrow, backspace, and delete work as if you were editing in a word processor (i.e. as normal)

Note that there's some variability in the expected behavior to implement this feature. This represents what I consider to strike a good balance between easy to implement/maintain, and what feels intuitive (to me). Feel free to suggest alternatives in how tags are selected/deselected, or how selected tags are drawn. 


jrozner:
Changes look good from my end. @jkennedyvz is going to talk with reporters for final feedback

PR #41, opened 2020-08-11
Remove syntax highlighting for markdown renders

JoelAtDeluxe:
This PR removes the feature to render codeblocks in markdown. Code blocks within markdown are still supported, but are rendered as if they were not in a code block.

i.e. markdown that looks like this:
```md
   ```java
      public static void main(String[] args) {
          System.out.println("Yo!");
      }
   ``
```
will look like:
public static void main(String[] args) {
&nbsp;&nbsp;System.out.println("Yo!");
}

This has been done to improve the responsiveness of loading markdown descriptions, which generally don't contain codeblocks in the first place. 


jkennedyvz:
Hi @JoelAtDeluxe ,

Thanks for this optimization.

PR #42, opened 2020-08-14
Add common database seed for unit tests as well as development data

JoelAtDeluxe:
This PR makes the (services) unit-test database seeding work for dev users. This is done because the unit test data is, generally, more robust than the dev data. In addition, this PR bulks up the test data by adding new users and content. All content is real, including a mini terminal recording. In addition, moving this seeding to a common area allows other packages to re-use seeding rather than generate their own.

Note that this solution attempts to modify as few files as possible, which ultimately leads to a large re-wrap file in the services package. I am opting to leave this rewrap for this PR, and, if necessary, begin replacing that rewrap later, as needed, as well as a separate PR adding or replacing existing testing with new seed data.


JoelAtDeluxe:
1. Yeah, actually it was intentional to change from email address to first name. I did this because I wanted easier to type and remember usernames and passwords, and to remove the expectation that user keys must be email addresses. I'll add a note on testing with the dev data.
2. That's odd. I'm not experiencing the same thing locally. This suggests that the file did not get copied into the container. Everything looks good on my end, but I'll double check with a fresh checkout of the branch

PR #58, opened 2020-09-04
Cache Busting Revision

JoelAtDeluxe:
This PR makes a change that should hopefully make cache busting consistently work.

The core of this change is by dynamically generating the index.html file with dynamically named main.js and main.css files. This should prevent errant loading of old assets, as each should have a new, unique name every time the service is built and deployed. 
This is accomplished by adding two libraries that interact with webpack. The first, `html-webpack-plugin`, generates the dynamic index.html file and stores this in memory. The second, `html-webpack-harddisk-plugin` allows us to write this dynamic html file to disk at the specified location.
In addition, this also updates the `nginx.conf` file to attempt to do as much as possible to prevent the index.html file from being cached itself. All other files (located under `/assets/`) should have normal caching rules applied to them.

In addition, a handful of small, but related changes, have been made to make testing easier. The first is restoring (And adding) the old seed data, to allow the production docker-compose file to provide a pre-seeded database. This should be amended in the future, but is far outside the scope of this PR, hence the easy solution of restoring the old content (note: this only restores the data, not the content). The second is simply an easy way to start up the production docker-compose containers via the makefile.

### DIFF
--- a/frontend/webpack.config.js
+++ b/frontend/webpack.config.js
@@ -1,4 +1,6 @@
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
+const HtmlWebpackPlugin = require("html-webpack-plugin");
+const HtmlWebpackHarddiskPlugin = require("html-webpack-harddisk-plugin");
 const path = require('path')
 
 module.exports = (env, argv) => ({
@@ -6,7 +8,7 @@ module.exports = (env, argv) => ({
 
   output: {
     publicPath: '/assets/',
-    filename: 'main.js',
+    filename: 'main-[hash].js',
     chunkFilename: '[chunkhash].js',
   },
 
@@ -39,8 +41,15 @@ module.exports = (env, argv) => ({
   },
 
   plugins: [
+    new HtmlWebpackPlugin({
+      title: "ASHIRT",
+      alwaysWriteToDisk: true,
+    }),
+    new HtmlWebpackHarddiskPlugin({
+      outputPath: path.resolve(__dirname, 'public')
+    }),
     new MiniCssExtractPlugin({
-      filename: 'main.css',
+      filename: 'main-[hash].css',
       chunkFilename: '[chunkhash].css',
     }),
   ],
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,11 @@ run-backend: clean-backend rerun-backend
 rerun-backend:
 	docker-compose up --build
 
+.PHONY: run-prod
+run-prod:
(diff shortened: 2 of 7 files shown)


jkennedyvz (approved):
Hi Joel,

These changes look fine. Thanks for adding the new seed data!

frontend/webpack.config.js
```diff
@@ -1,12 +1,14 @@
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
+const HtmlWebpackPlugin = require("html-webpack-plugin");
```
jrozner, line 2: Do we actually need to break out all the CSS files? Won't we get better performance leaving it bundled?

jrozner (changes requested):
Not opposed to these changes but curious about it. Do we know the root cause of the issue? Was the index.html not being cache busted? Was the main.css/js not being cache busted and attempted to load other older files that no longer exist? Does adding the hash to the main files fix the problem alone? I'd like to avoid turning off caching for index.html if we can avoid it and avoid people needing these specific configs if they were to deploy without the container and our nginx config.

JoelAtDeluxe:
We think what's happening is that the main.js file is being cached (given that the index.html previously wasn't changing, I don't know that it necessarily mattered if the index.html was cached or not). The main.js etag was not consistently being updated after each new build. This in turn could load the wrong js chunks on subsequent requests.  It was unclear why nginx was not correctly generating a new etag -- the criteria seems to be based on the creation date and file size, so both of those should really trigger a new etag on each new build.

As for the main.css, we hashed this because we didn't want to end up in the same situation we were (maybe) seeing with the main.js file. Likewise, we had some mild concerns that even the index.html file could be improperly cached, and trigger the same issue, so that's why we removed the cache on the index.html file.

frontend/webpack.config.js
JoelAtDeluxe: Just a quick note that this isn't a new change -- it's been in master probably next to forever. 
As to whether or not a single file would lead to better performance, I'm not sure. I'll read up to see if there's preferred practice for this.

jrozner: Ah, didn't realize this wasn't new. Don't worry about it then.

JoelAtDeluxe: Just for clarity, we do the css splitting because it does improve the first load. The common styling is included in the main.css file. Specific css is loaded as needed. A single css would be better if the javascript were delivered as a single file.

jrozner (approved):

PR #59, opened 2020-09-08
Try to improve description rendering

JoelAtDeluxe:
This PR attempts to further speed up description rendering on the timeline. 

The current issue faced is that rendering markdown is slow due to a bottleneck with asset loading. Descriptions are rendered in markdown and the markdown renderer is dynamically loaded. This means that when the other dynamic elements come in, the code needs to react to both downloading those assets, as well as downloading the markdown rendering code.

This change simply preloads the markdown renderer, which should show all descriptions as soon as the timeline component loads, as well as the request to get evidence completes. There should no longer be a situation where the loading animation plays where the description is supposed to be.

This affects all areas where the markdown renderer is used (evidence descriptions, and finding descriptions). 

### DIFF
--- a/frontend/src/components/markdown_renderer/index.tsx
+++ b/frontend/src/components/markdown_renderer/index.tsx
@@ -2,19 +2,14 @@
 // Licensed under the terms of the MIT. See LICENSE file in project root for terms.
 
 import * as React from 'react'
-import {useAsyncComponent} from 'src/helpers'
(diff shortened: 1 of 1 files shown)


jrozner (approved):
Don't have a really easy way to test this without deploying because we don't have real latency or enough real images to block locally. Changes seem reasonable and there is no downside. Let's merge this and see if it helps. We can always revert if needed

PR #60, opened 2020-09-21
Add API check connection endpoint

JoelAtDeluxe:
This PR adds a simple connection check endpoint to the api server. This is done to provide a more robust way to ensure that an api is talking to an actual ASHIRT server. 

In psuedo code, the recommended solution for verifying the server is:

```python
def check_connection(host: string, accessKey: string, secretKey: byte[]) -> string:
  resp = network.GET(apiHome() + "/checkconnection", {
     "Authorization": accessKey + ":" +secretKey,
     "Date": get_rfc_1123_date()
  })
  if resp.status == HTTPStatus.OK:
    parsedData = json.Parse(resp.data)
    if parsedData.parsedSuccessfully() and parsedData.ashirtConnected == true:
      return "Connected"
    else:
      return "Incorrect or outdated server"
  elif resp.status == HTTPStatus.Unauthorized:  # http code: 401 
    return "Access or Secret Key mismatch"
  elif resp.status == HTTPStatus.NotFound:  # http code: 404
    return "Server not present at address"  # Check host string
  else:
    return "Unexpected error: " + resp.status  
```

### DIFF
--- a/backend/server/api.go
+++ b/backend/server/api.go
@@ -13,6 +13,7 @@ import (
 	"github.com/theparanoids/ashirt-server/backend"
 	"github.com/theparanoids/ashirt-server/backend/contentstore"
 	"github.com/theparanoids/ashirt-server/backend/database"
+	"github.com/theparanoids/ashirt-server/backend/dtos"
 	"github.com/theparanoids/ashirt-server/backend/logging"
 	"github.com/theparanoids/ashirt-server/backend/models"
 	"github.com/theparanoids/ashirt-server/backend/server/middleware"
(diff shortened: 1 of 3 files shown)


backend/server/api.go
```diff
@@ -51,6 +51,13 @@ func bindAPIRoutes(r *mux.Router, db *database.Connection, contentStore contents
 		return services.ListOperations(r.Context(), db)
 	}))
 
+	route(r, "GET", "/api/checkconnection", jsonHandler(func(r *http.Request) (interface{}, error) {
+		reply := struct {
+			Connected bool `json:"ashirtConnected"`
```
jrozner, line 56: Can we change this to 
```go
Ok bool `json:"ok"`
```

jrozner (changes requested):

backend/server/api.go
JoelAtDeluxe: We can. my motivation here was to make sure we were connecting to an ashirt server, but this kind of issue seems rare anyway. Parsing Ok only should be fine.

jrozner (approved):

Issue #75, opened 2020-09-29
Account recovery is broken


JoelAtDeluxe:
**Golang version** (`go version`): 1.14

**OS version** (`uname -a` if on a Unix-like system): N/A

**Description of the problem including expected versus actual behavior**:

When a recovery code/url is used, recovery validation occurs (and is successful), but the next call to /user fails due to an authentication issue, kicking the user back to the login screen.

**Steps to reproduce**:

 1. Have an instance with 2+ users: one admin (call 'A') and one other (call 'B')
 2. Have the admin log in. Go to the admin screen, then user management
 3. Find user B, then click the actions arrow and select "Generate Recovery Code". Copy the recovery code, and in an incognito/private browser window, try using that code
 
Result: user gets sent to the autherror/recoveryfailed endpoint, and the user can only return to the main menu
Expected Result: the user should be logged in temporarily, and be prompted to change their password.

**Provide logs (if relevant)**:
```
timestamp=2020-09-29T22:36:12.505346177Z ctx=e4fa9531-1234-444d-9e79-88070c360fa9 msg="Incoming request" method=GET url="/auth/recovery/login?code=usLFdafsSK_RMb0a0T_e-oY-XRhuWTIgiPMWsRNQApRunTkFPqfquA==" from=192.168.144.4:47614
timestamp=2020-09-29T22:36:12.50575902Z msg="executing query" query="SELECT user_id FROM auth_scheme_data WHERE user_key = ? AND TIMESTAMPDIFF(minute, created_at, ?) < ?" values="[usLFdafsSK_RMb0a0T_e-oY-XRhuWTIgiPMWsRNQApRunTkFPqfquA== 2020-09-29 22:36:12.505709587 +0000 UTC m=+19.113118880 1440]"
timestamp=2020-09-29T22:36:12.506491846Z msg="executing query" query="DELETE FROM auth_scheme_data WHERE user_key = ?" values="[usLFdafsSK_RMb0a0T_e-oY-XRhuWTIgiPMWsRNQApRunTkFPqfquA==]"
timestamp=2020-09-29T22:36:12.517034011Z ctx=e4fa9531-1234-444d-9e79-88070c360fa9 msg="Request Completed" status=302 sizeInBytes=48 duration=11.689908ms
timestamp=2020-09-29T22:36:12.748556113Z ctx=4a2cfeb4-1034-499b-97db-a562fe8ccfa0 msg="Incoming request" method=GET url=/user from=192.168.144.4:47632
[...]

PR #76, opened 2020-09-29
Fix recovery auth

JoelAtDeluxe:
This PR corrects an issue where every One-Time Auth would fail, due to how error wrapping works. This also introduces a unique error page for admins that create a recovery code and try to use it while logged in (Technically this happens for any logged in user, but in practice this should only happen for admins that try to verify recovery urls)

Addresses PR #75 


jrozner (approved):
Verified fix

PR #88, opened 2020-10-01
Add in 2-factor authentication via TOTP

JoelAtDeluxe:
This PR adds in the ability for users to add a 2nd factor authentication via TOTP (time-based one-time passwords) for local authentication (only). 

Changes:
* Add TOTP check in local auth
* Update database to include totp field in auth_scheme_data table
* 5 new TOTP-related api methods for local auth (effectively: hasTotp, generateTotp, registerTotp, finishLoginWithTotp, removeTotp) -- note, admins have access to query user's totp status and remove totp from a user account
* Users with TOTP enabled will now reach a screen where they must provide TOTP details before they can continue. 
* New security section to account settings to add/remove TOTP to a user account
* Admins can now remove TOTP for a user via admin's User Management, specifically with the action "Remove Two-factor authentication" (includes confirmation prompt)


jrozner:
It looks like there is a bug in the template here. It's dropping the string of the object rather than the path itself.
Screen Shot 2020-10-08 at 9 32 10 AM

JoelAtDeluxe:
As noted in a separate conversation, the missing icon is actually due to an earlier merged  PR that updated file-loader. A separate PR will fix this issue (PR #112 )

jrozner (changes requested):
Functionality looks good. Can we update places where it says "two-factor", "two factor", etc. to be "multi-factor"?

jrozner (approved):

PR #212, opened 2021-01-08
Support HTTP request cycle Evidence (HAR viewer)

JoelAtDeluxe:
This PR adds  the abiilty for users to upload and view HAR files (other  formats can be supported via transformation  processes in the backend).

Changes:
* Minor backend changes to add a new content type
* New frontend component to render a series of http request/responses (similar to what you would get inside chrome or firefox)
* Added the ability for tables to have selected rows (via implementing a onKeyDown  function and providing a ref for the table)


JoelAtDeluxe:
In general, it looks like this design is going to require approximately 950ish pixels for the page content in order to look okay. The easiest change to expand this limit is going to be to change the maximum height for the table, and adjust it downward to only 4 rows (including the header) are shown at any one time -- that'll net about 40ish pixels. 

We might also need to consider a different layout -- either completely different in how this is laid out, or some structural changes to make the component render better in the timeline view.

jrozner:
@JoelAtDeluxe can you get this rebased? Where are we at in terms of re-design? Is this related to the min-width or min-height needed currently for this to work?

JoelAtDeluxe:
I'll rebase this tomorrow. Currently, the design "wants" about 950 pixels, but it can grow a bit. If you've got a 4k screen, it might not look great.

However, I don't think I ever looked at the add evidence view (from within a finding). This view is very small, and so this design is probably going to be completely broken there. I'll need to think on that, in addition to just trying to make it scale with page height better. I'll look around a bit more to see how others implement this sort of thing.

jkennedyvz (dismissed):
Looks great, thanks for working through all the sizing issues.

jrozner (approved):
Looks good. Still not totally happy with all the the UI, especially as we go down smaller but we can iterate on this.

jrozner:
Are we still waiting on anything here or are we good to go after we fix the conflict? I think we'll probably need to take another pass at some point at some of the and hotkeys but I think in general we probably need to do that for the whole app around hotkeys, modals, and some of the ui elements.

jkennedyvz:
This is ready, it just needs to be rebased.

JoelAtDeluxe:
Sorry, I didn't know this got stuck with a conflict. I can resolve that momentarily

jkennedyvz (approved):

PR #218, opened 2021-01-15
Implement ordered results for timeline queries

JoelAtDeluxe:
This PR adds the ability to order the timeline results in an ascending or descending (current, default) order, by time, when adding a new parameter to the search query.

Users may now add to the search the following:
`sort-direction:asc` to order in an ascending fashion (first-to-happen view). Instead of `asc`, you may provide synonyms: `chronological` or `ascending`. All of these values are case insensitive
Any other value (or not specifying this field) returns the same results in a descending fashion (most recently occurred first)

This can be  used on any timeline view, and will kept for saved queries.

Note: Currently, all of the mock data occurs at the same time, so the sorting will not be obvious for these records -- add new evidence to test this feature.

### DIFF
--- a/backend/helpers/query_parser.go
+++ b/backend/helpers/query_parser.go
@@ -28,6 +28,7 @@ type TimelineFilters struct {
 	DateRange        *DateRange
 	WithEvidenceUUID string
 	Linked           *bool
+	SortAsc          bool
 }
 
 // ParseTimelineQuery parses a query a user may type into the search box on the timeline page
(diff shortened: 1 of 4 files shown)


jkennedyvz (commented):
Hi @JoelAtDeluxe 

The changes work as expected. The option feels very verbose when I type it, is there a way to alias something like "sort:asc"?

JoelAtDeluxe:
Yeah, we can change the name. I wasn't sure what would be the best solution, so opted for the most descriptive. I'll switch it to `sort` for now, and look into how reasonable adding a longer but more descriptive alias

jkennedyvz (approved):
I'm happy, thanks @JoelAtDeluxe !

jkennedyvz (approved):

jkennedyvz (approved):

jrozner (approved):

PR #219, opened 2021-01-18
Add help text for search bar

JoelAtDeluxe:
Adds help text for the search bar. The help icon is the question mark located _inside_ the search text bar. Clicking the icon will open a modal with a lengthy amount of help text to inform how to use the filtering features for  searching the timeline.


jkennedyvz (approved):
Hi @JoelAtDeluxe 

This help box looks great, thanks for the addition.

jrozner (changes requested):
1. It looks like the esc key doesn't close the modal like the others, can we fix that?
2. This isn't super important but the evidence modal can pop up over the help modal and I've noticed this sort of behavior across other modals. Is there a consistent ordering of which modals can pop over which? Does it make sense revisit this later?
3. Is there a specific ordering currently for the various options? Should we alphabetize them?

JoelAtDeluxe:
> 1. It looks like the esc key doesn't close the modal like the others, can we fix that?
Oops. Yes, I'll fix that.

> 2. This isn't super important but the evidence modal can pop up over the help modal and I've noticed this sort of behavior across other modals. Is there a consistent ordering of which modals can pop over which? Does it make sense revisit this later?
There is no defined order. What happens is that every modal gets appended to the DOM, and whatever is last is on top. We can force an order if this is a problem. 

> 3. Is there a specific ordering currently for the various options? Should we alphabetize them?
No specific ordering -- just matches what we have on the go side, which would be roughly in order of when the filter capabilities were expanded. I don't know if alphabetical is the right order here, as it kind of hides (what I presume are) more commonly used filter options. I'll take a stab at what I think is the most used ordering, but I'm happy to switch this to alphabetical if you think that'll be easier or more maintainable.

| Alphabetical | Joel's Suggestion |
| --- | --- |
| linked     | tag |
| operator | operator |
| range     | range |
| sort        | sort |
| tag         | linked |
| uuid        | with-evidence|
| with-evidence | uuid |

Also, looking at this list, we might want to introduce a new filter option: content type

jrozner (approved):

jrozner:
Changes look good

PR #223, opened 2021-01-28
Add search builder modal

JoelAtDeluxe:
This PR creates a modal for aiding in constructing a search string. ~All~ Almost all of the current search fields are supported, and filtering specific to evidence timelines or finding timelines are also hidden away when not appropriate.

One limitation here is the `uuid` filter, which restricts the output to a single evidence or finding. I did not think of a good way to represent this, and it also makes all other filters pointless, so I simply left that out. If a user does specify the `uuid` flag for the input, then it will be silently passed through to the resulting output string -- there just won't be a way to edit it in the search builder form.


JoelAtDeluxe:
One current issue exists with this PR: when adding or updating a findings saved query, the browse  for evidence button doesn't entirely work. This ends up being a form-in-a-form, where the submit button on the inner form  propagates  down to  the edit  form, which ultimately closes the modal. I'm still working on this scenario, but everything else seems to be working.

jrozner (approved):

jkennedyvz (approved):

PR #248, opened 2021-02-11
Log out user after their system state changes

JoelAtDeluxe:
This PR provides the ability to log  out a user when that  user's  "user flags" are changed. Specifically, when a user moves from a normal user to either a disabled user/admin user, or from one of those states back to a normal user (or really any combination of admin/disabled flag changes).

This fixes a bug and restores previous functionality that had been lost along the way.

### DIFF
--- a/backend/session/store.go
+++ b/backend/session/store.go
@@ -69,7 +69,14 @@ func (store *Store) readRaw(r *http.Request) *sessions.Session {
 func userIDFromSession(sess *sessions.Session) *int64 {
 	sessionData, ok := sess.Values[sessionDataKey].(*Session)
 	if ok {
-		return &sessionData.UserID
+		val := sessionData.UserID
+		// userIDs start at 1. In some cases, the UserID won't be set and will be 0
+		// trying to add a row with userID 0 will result in a foreign key constraint error, so
+		// returning nil ("no user id") instead
+		if val != 0 {
+			return &val
+		}
+		return nil
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


jkennedyvz:
Hi @JoelAtDeluxe,

This functionally works, however I am not returned to the login page in some cases after logout. After the session is invalidated I can still click around the user interface. Is there a simple way to fix this?

JoelAtDeluxe:
There's no easy way to redirect the user back to the index page after they've been logged out from the backend. 

Both the frontend and the backend have some sense of who the user is when they're logged in. When the user logs out from the frontend, they clear their context, and tell the backend to do the same, then jump to the login page. Since both the frontend and the backend are in sync, everything looks good. When a user is forced to log out (via the backend), the backend clears it's context, but cannot tell the frontend to to the same. The frontend needs some way to determine if a user is still logged in. While we have that mechanism, there's no convenient place to do so on the frontend. If we really want to do this, we'll probably need to implement this on each page, and maybe even on particular components, depending on exactly how we want this to work.

backend/session/store.go
```diff
@@ -64,3 +65,18 @@ func (store *Store) readRaw(r *http.Request) *sessions.Session {
 	sess, _ := store.wrappedStore.Get(r, "auth") // ignoring, because errors only fire if a bad cookie name is provided
 	return sess
 }
+
+func userIDFromSession(sess *sessions.Session) *int64 {
+	sessionData, ok := sess.Values[sessionDataKey].(*Session)
+	if ok {
+		val := sessionData.UserID
```
jrozner, line 72: This all makes sense. Is there a reason that we didn't make the UserID in sessionData nullable and just pass it through so we don't end up with the default 0 instead of nil?

JoelAtDeluxe: I honestly can't remember why these were set to not null.

jrozner (approved):

PR #253, opened 2021-02-20
Self-Service Account Recovery

JoelAtDeluxe:
This PR adds the ability for a user to request a recovery url delivered to the email on file. This only really applies for local auth.

Changes: 
* Add a request  recovery page, and a link to that on the main page
* Updates to the recover auth to have the ability to generate a code on behalf of the requesting user 
* Add a database table to keep track of the emails to be sent
* Add in a worker to check for unsent emails
* Add in an email servicer, which sends the actual email. Three structs implement the interface (two for testing, one for real life)
* Add in configuration options for email servicer

TODO:
* implement rate limiting for the email servicer.


jkennedyvz:
Hi @JoelAtDeluxe, when the email recovery is used we would like the user to be taken to the /account/authmethods page. 
In the email template we should include instructions for how the user should delete and re-link Okta or local authentication. When the user re-links local auth it allows them to reset their password without knowing the previous one. 

This existing method of re-linking auth could be abused for full account takeover if a user session is stolen is leaked. We should consider adding a control around the auth methods that requires either the user password (which an admin could reset), or some other sort of admin-generated token to authenticate the user.

jkennedyvz (approved):

PR #284, opened 2021-03-31
Allow for closing local registration, allow admin to create new local users

JoelAtDeluxe:
This PR does two things: First, it adds a new startup flag that will disable local registration. Second, it allows admins to create new user accounts for users on their behalf. Documentation has been provided to indicate how to turn on this feature, and some revised instructions has been provided on how to register the first account (especially when local auth registration is blocked)

### DIFF
--- a/backend/authschemes/localauth/local_auth.go
+++ b/backend/authschemes/localauth/local_auth.go
@@ -106,7 +106,7 @@ func (p LocalAuthScheme) BindRoutes(r *mux.Router, bridge authschemes.AShirtAuth
 		if _, err := rand.Read(authKey); err != nil {
 			return nil, backend.WrapError("Unable to generate random new user password key", err)
 		}
-		
+
 		// convert authKey into readable format
 		bufKey := make([]byte, len(authKey))
 		for i, b := range authKey {
--- a/backend/authschemes/localauth/local_auth.go
+++ b/backend/authschemes/localauth/local_auth.go
@@ -108,11 +108,7 @@ func (p LocalAuthScheme) BindRoutes(r *mux.Router, bridge authschemes.AShirtAuth
 		}
 
 		// convert authKey into readable format
-		bufKey := make([]byte, len(authKey))
-		for i, b := range authKey {
-			bufKey[i] = (b % (126 - 33)) + 33
-		}
-		readKey := base64.StdEncoding.EncodeToString(bufKey)
+		readKey := base64.StdEncoding.EncodeToString(authKey)
 
 		dr := remux.DissectJSONRequest(r)
 		info := RegistrationInfo{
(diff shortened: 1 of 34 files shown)


backend/authschemes/localauth/local_auth.go
```diff
@@ -56,39 +73,68 @@ func (LocalAuthScheme) FriendlyName() string {
...
-		})
-		if err != nil {
+		authKey := make([]byte, 40)
+		if _, err := rand.Read(authKey); err != nil {
+			return nil, backend.WrapError("Unable to generate random new user password key", err)
+		}
+		// convert authKey into readable range (33-126)
+		readKey := make([]byte, len(authKey))
+		for i, b := range authKey {
+			readKey[i] = (b % (126 - 33)) + 33
```
jrozner, line 111: Why are we using mod to get the value into the ascii printable keyspace rather than just encoding to something printable (hex or base64)? Isn't this reducing the keyspace?

jrozner (changes requested):
Implementation looks pretty good. I think we're probably want to revisit this along with account recovery, and admin password resets once first versions of everything is in and figure how what makes sense.

I think in the long run it might be nice to just email the user the temporary password or a single time use login link (like for account recovery) to allow the user to login. Depending on implementation maybe if we go this route we don't create local auth for them and allow them to choose an auth scheme to link, or we just keep the local auth as is.

backend/authschemes/localauth/local_auth.go
JoelAtDeluxe: My thinking at the time was that it was important to cast these values into a printable range, and having a larger set range of characters was better than going with base64 (after all: this range is 93 possible characters, rather than the 64 we'd get otherwise)

But thinking about this now, I think that's flawed. I think the entropy should be the same for the 40 characters represented as base64 as they would be in any other representation. Also, I'm not confident that the above mapping maps equally over all of the 93 possible characters, so the base64 approach might be better after all. I can revise into that approach if you agree.

backend/authschemes/localauth/local_auth.go
```diff
@@ -56,39 +74,70 @@ func (LocalAuthScheme) FriendlyName() string {
...
-		})
-		if err != nil {
+		authKey := make([]byte, 42)
+		if _, err := rand.Read(authKey); err != nil {
+			return nil, backend.WrapError("Unable to generate random new user password key", err)
+		}
+
+		// convert authKey into readable format
+		bufKey := make([]byte, len(authKey))
+		for i, b := range authKey {
```
jrozner, line 112: If we're using base64 we don't need to do the modulo math here to get into an ascii printable space because base64 is always printable. This still reduces the keyspace because the input going into base64 is constrained to just the printable space already.

jrozner (changes requested):

backend/authschemes/localauth/local_auth.go
JoelAtDeluxe: Oh, sorry, i meant to remove that. I guess I was too eager with my commit.

jrozner (approved):
Looks good

PR #297, opened 2021-04-08
Add support for  dynamic categories

JoelAtDeluxe:
This PR allows for admins to define and edit the categories that can be assigned to a finding.

Changes:
 * Added a new table to track the supported categories
   * This field supports deletes-in-name, where a deleted_at column determines whether this category is visible generally. This  will provide the ability for findings to maintain their original category even after a category is deleted.
 * Added endpoints to list, create, update and delete/restore (this endpoint does double duty)
 * Added a new section in the admin tools operations section which will allow creating/editing/deleting finding categories

Note: this is a partial implementation. A second PR should be implemented that would convert the findings table `category` column to instead be a category_id column, pointing to the `finding_categories` table. This helps solve an issue where a finding_category and a _finding's_ category can drift over time. This is a non trivial change, so this PR limits the change strictly to making the categories dynamic, and will address the above in a separate PR.


jrozner (changes requested):
Changes look pretty good. Two requests:
1. Can we get a new section in the left rail for the finding categories so it's on it's own page?
2. Is it difficult to inform the user of how many findings are using the category that's being deleted in the confirmation modal?

In terms of deleted categories, is there any way to completely from them from the list? I could see it getting a bit messy if the list just grows over time.

JoelAtDeluxe:
Both of those changes are fairly easy. 
1. Still working on this, but it's pretty close to trivial. I don't expect any serious changes to any behavior.
2. All we really need to do this is change the query to also join findings and find usages. It's a _slightly_ complicated query, but we already have something like it with tags.  From a user perspective, when they choose delete, they'll get one of three messages:
  1. If no findings are affected: "Are you sure you want to delete this finding category?" 
  2. If only one finding is affected: "Are you sure you want to delete this finding category? This will affect one finding."
  3. If more than one finding is affected: "Are you sure you want to delete this finding category? This will affect ${number_of_affected_findings} findings."

Obviously the text can change, and there might be some room for some functionality changes, but in general:  yes. When we get the list of finding categories, we'll have the number of usage _at that instant_. Does the above sound appropriate, or do you want to change that?

JoelAtDeluxe:
Oh, I missed the last point. Yes, I can either always hide the deleted findings, and never allow a restore, or show a checkbox that changes what is shown.

JoelAtDeluxe:
All of those requests should now be addressed. I opted to show a checkbox, and it will always default to unchecked when that page loads. Clicking it will show the deleted categories.

jrozner (changes requested):
This looks great. One more small change, can we get the number of findings that are using the category directly in the table?

jrozner:
Looks good

jrozner (approved):

PR #325, opened 2021-05-26
Fix local auth linking over user email

JoelAtDeluxe:
This PR address an issue where a user could delete their local auth account, then re-link by supplying an email address currently used by a user. (an email column exists both as a user column and as an auth column. For auth, it's what you would use to login. For user, it's more of a contact thing)

Note: this same fix has been added to the `js-self-service-recovery` branch,  so a conflict is likely when one of these is merged in.

### DIFF
--- a/backend/authschemes/auth_bridge.go
+++ b/backend/authschemes/auth_bridge.go
@@ -5,12 +5,14 @@ package authschemes
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


JoelAtDeluxe:
Closing, as this was handled in the self-service-recovery PR

Issue #368, opened 2021-08-04
Group operations by status on main page.


legitimatecoder:
Once you have completed many operations, the main page can get very busy, with lots of completed operations filling up the page and making it harder to find a specific one still being worked on.

It would be cool to have the pending/active/complete operations in separate sections/groups for a better visual context of which ones are still being worked on.

Also, having the ability to toggle hide/collapse those groups would be useful for managing the visual clutter.

jkennedyvz:
I've noticed my own workspace becoming a bit cluttered, great suggestion.

PR #430, opened 2022-01-13
add a filter input to help find operations

JoelAtDeluxe:
This PR adds a filter input to the operations list page (where you can create a new operation). This is to help solve issue #368. While this isn't a direct solution to that issue, it does provide a mechanism to achieve the same end, while also not "hiding" an operation by moving it into a new area. We can extend this feature later if needed to include additional filtering mechanisms, or we can abandon this to move to a grouping approach, if that's generally more helpful.


legitimatecoder:
Yeah, this is an acceptable work-around. Thanks!

jkennedyvz (approved):

PR #435, opened 2022-01-17
GCP backed storage

JoelAtDeluxe:
This PR adds the ability to store files to GCP.

-- Under Construction --


jkennedyvz:
Would it make sense to have the development stack write the seed data to the storage provider? We should also include the new storage variables in the compose files.

JoelAtDeluxe:
We should be able to seed the storage provider as well. I'll take a stab at it.

jkennedyvz:
Thanks @JoelAtDeluxe , one other request - I don't see `GOOGLE_APPLICATION_CREDENTIALS` in the readme or example compose, only the PR notes. Can we add those in?

jkennedyvz:
Looks grea, I was able to test content seeding with GCP and AWS!

jkennedyvz (approved):

PR #437, opened 2022-01-24
Add support for event-type evidence (Demo)

JoelAtDeluxe:
This PR takes a stab at creating event evidence. This type of evidence contains no content, but does have a placeholder image to help signify that an event occurred.


JoelAtDeluxe:
Marking this as a draft PR, but it's more or less done based on what I believe to be correct based on the ticket.

jrozner:
This seems good. I'll look around for an icon we can use for the event to replace the image there but otherwise this looks good to go.

jrozner (approved):
Changes look good.

jrozner (approved):

PR #438, opened 2022-01-25
Add Copy Evidence button

JoelAtDeluxe:
This PR adds a button overlay on some evidence that enables the ability to copy the core content of that evidence (the image for screenshots, the source code for codeblocks). This ability theoretically should work cross browser and cross OS, but is limited in a number of ways. Text copying should always work, while image copying will work depending on both the storage type of the image and the browser being used, as well as how that browser is configured. Below is a short guide on what to expect:

### Using Chrome

Text will copy and png images should copy. All of types of image files will not copy. You may get a modal indicating the failure to copy

### Using Firefox

Text will copy. PNGs may copy if the correct flag is turned on. Other file types will not copy. You may get a modal indicating the failure to copy

To enable the correct firefox flag:
* Open a new tab, and navigate to `about:config`
* Look for the `dom.events.asyncClipboard.clipboardItem`. Set this value to **true**. Note that you'll also need `dom.events.asyncClipboard` also set to true, but this is enabled by default
* Note that this enables this flag universally

### Using Safari

Still testing, but it might work for everything.

Please note that in all cases, you should be able to right click and choose "save image" which should work regardless of the image type/OS


jrozner:
Based on limitations and the fact that you can just right click and copy the image contents, which was the main use case, maybe we just hold off on this for now. I think there is still maybe value in the feature, especially for other evidence types (eg. storing a har or asciinema file on the clipboard as a data uri to bring into something else) but that seems like it's not worth tackling now.

In terms of feedback, it seems to work pretty well. I think if we re-visit this I'd like to see the option folded into the dropdown where move and delete currently are and have it disabled if it's not possible to copy that specific evidence type.

It also may make sense to start storing the mime type of the evidence for use in determining whether the evidence could be copied or not. We may also want that in the future with the enrichment pipeline.

PR #460, opened 2022-02-12
Add support for initial tags when creating a new operation

JoelAtDeluxe:
This PR adds the ability for admins to define a set of default tags that will be copied whenever a new operation is created. These are true copies, so operation owners can tweak these to fit their needs without interfering with the default tags.

This functionality is implemented in the admin "Operation Management" section. 


jkennedyvz (approved):

jrozner:
This looks good. Can we move the default tag management out onto it's own page in the admin section rather than putting it in operation management?

JoelAtDeluxe:
Yeah, that's an easy change.

JoelAtDeluxe:
Just a note that the requested change was complete, and ready for re-review

jkennedyvz (approved):

PR #471, opened 2022-02-22
remove access to operations via ID

JoelAtDeluxe:
This PR removes the outdated code and endpoints to access operations by ID (instead of by slug).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ test-all: test-backend test-frontend
 
 .PHONY: test-backend
 test-backend:
-	cd backend && go test ./...
+	cd backend && go vet ./... && go test ./...
 
 .PHONY: test-frontend
 test-frontend:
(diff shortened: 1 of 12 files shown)


jrozner:
We should also be able to remove https://github.com/theparanoids/ashirt-server/blob/master/backend/dtos/dtos.go#L47-L50. Do you know if ashirt or aterm still use this? Do we need to make any change to them?

JoelAtDeluxe:
Oh, forgot about this. No, those two both use slugs only in their network requests. I don't think they would use the IDs for any other purposes. I'll add that bit to this PR, and check out those two a bit closer to see if IDs are ever used for operations.

JoelAtDeluxe:
We need to make a minor change to ashirt before we can merge this change. When we parse operations in the ashirt app, we parse (but don't use) the id field. So, we need to remove that, and then we should be good to go. I'll open up a PR for that soon.

JoelAtDeluxe:
As a note, this PR should only be merged once this minor AShirt PR is merged:
https://github.com/theparanoids/ashirt/pull/112

jkennedyvz (approved):

PR #473, opened 2022-02-25
Support importing/exporting default tags

JoelAtDeluxe:
This PR adds the ability to share default tags. The sharing format is provided as a readable json format, allowing for hopefully easy editing. Note that this shares _all_ tags and leaves editing up to the end user, for better or worse.


jkennedyvz:
The only issue I found is that I can import a tag with no name, and this is disallowed in the normal tag creation flow. The merging and duplication checks all work well. 

Screen Shot 2022-03-01 at 3 12 39 PM

JoelAtDeluxe:
Good catch. I'll strip out blank entries and just accept everything that's not blank.

jkennedyvz (approved):

PR #485, opened 2022-03-02
Restore default tag import/export

JoelAtDeluxe:
This PR just brings the import/export feature back to a place where it can be merged


jrozner:
Can we do file import/export rather than having to copy/paste?

jkennedyvz:
In chrome I am getting 
`Download is disallowed. The frame initiating or instantiating the download is sandboxed, but the flag ‘allow-downloads’ is not set. See https://www.chromestatus.com/feature/5706745674465280 for more details.`

jkennedyvz (approved):

PR #490, opened 2022-03-14
Allow for key rotation

JoelAtDeluxe:
This PR implements a solution to rotate an api key for a user. Simply go to the API key management section (`$ASHIRT_DOMAIN/account/apikeys`), and next to delete, there's a new button to rotate, which ~will delete the current api key, and add a new one.~ Now revised to create a new api key, then allow for deleting the old one. This should help prevent momentary outages.

This PR also fixes an error where the api keys listed in that screen were always the calling user's API keys. This works in most cases, but for admins, when spoofing a user, the keys listed would be their keys, but all of the actions would target intended user. Now, users will always see the target user. If not an admin, they'll get an error when requesting keys for any other user.


jrozner:
I'm fine with landing this as is but I think we might want to re-think admin management of api keys and management of headless users. I don't know that this is super useful (except the bug fix that lists the correct keys). It doesn't seem like it makes anything particularly easier (clicking the rotate button and the button in the modal) rather than either creating a new keypair and migrating applications over, then deleting the old keypair or deleting the existing keypair first, creating a new one, and migrating applications over.

My understanding is that the real reason we wanted to implement something here was to make it easier to setup headless accounts. Correct? Or was there another reason?

I think what might make sense is to logically separate headless and non-headless users in the UI, since they are intended to be used totally differently, and maybe change around some of the interaction a bit between them. What might be really nice is to shortcut getting to api key management for a headless user but short of that editing the user, going to their api keys, and simply using the workflow we have now, seems fine.

Is it a big deal to split out the bug fix (and presumably tests) for the api keys fix that this includes, land that, and then think a bit more about what next steps might look like around supporting headless user management?

JoelAtDeluxe:
Closing because we want to rethink this feature.

PR #521, opened 2022-04-07
Update search UI

JoelAtDeluxe:
Work-in-progress branch on a possible UI change for how filtering works. Not at all ready to merge


jrozner:
In terms of functionality, this is much better since all the behavior works with the controls and there's no more issues with dropdowns not closing. From a ui standpoint I think it's a bit wide, especially when on a screen that you're pretty close to. At laptop distance, it's pretty hard to see the entire horizontal space at the same time so I constantly have to look back and forth. The original design it was easier to just go down the list of search critera. Maybe removing the 3rd column worth of controls would help?

The other nit pick, is it possible to make the button that drops down the search controls not move?
Screen Shot 2022-04-28 at 2 12 36 PM
Screen Shot 2022-04-28 at 2 12 46 PM

Also, can we see what it looks like with the pencil rather than the +?

PR #524, opened 2022-04-12
Add support for evidence metadata

JoelAtDeluxe:
This PR adds evidence metadata. The database seed now adds some default evidence metadata for playing, and I've added the ability to add and update metadata on the evidence timeline (this can be added to the findings timeline as well, but the word `metadata` is too big, so we need to decide on how it's placed there).

Edit: This PR is going to be re-worked into several smaller PRs for easier consumption


jrozner:
Overall this seems like a pretty good start. A few notes:
1. I think we can just put metadata under the dropdown menu for each evidence. I generally don't think we'll be actively diving into there on a per evidence basis unless we're debugging something or going to grab some valid specifically.
2. As of now, I think the UI implementation is generally pretty fine and a good enough starting point.
3. I'm still on the fence whether we'll want metadata to be editable by individual users. It might be something that we want to allow admins to do but given that the intention is that we want it to be handled by external applications, treating the information as immutable might make sense.

From a beyond the UI perspective we probably need to a think a bit about how things will work on the backend:
1. How we surface failures of a particular enrichment pipeline to users/admins
2. How we re-run enrichment for a particular evidence or all evidence in an operation. Do we wipe out existing data? Can you select just specific sources? etc.
3. Is a text blob for the metadata sufficient? How does that effect indexing and search? Do we need to do stemming or any sort of normalization or processing on the data before it goes in? If so, is that done by the api or the individual sources?

JoelAtDeluxe:
> Overall this seems like a pretty good start. A few notes:
> ...snip...
> 3. I'm still on the fence whether we'll want metadata to be editable by individual users. It might be something that we want to allow admins to do but given that the intention is that we want it to be handled by external applications, treating the information as immutable might make sense.

Yeah, I'm in a similar boat. It really depends on what metadata means. If it's just data extracted by some external workers, then editing doesn't make a lot of sense. Maybe in some odd edge case, but how we restrict that is certainly up for debate. But, if it's something that a user can create, then it seems rational to make it editable, and probably editable for anyone that can create/edit evidence. 

So, to me, the question is really more about what form the metadata takes. Also, it's easy enough to remove the edit button for any evidence that's deemed immutable. We just need to know what makes evidence immutable (e.g. it's from a tool)

> From a beyond the UI perspective we probably need to a think a bit about how things will work on the backend:
> 
> 1. How we surface failures of a particular enrichment pipeline to users/admins

Yeah, I'm thinking about this. It's more of a pipeline concern than a metadata concern, but there is a bit of overlap. I'm open to ideas.

> 2. How we re-run enrichment for a particular evidence or all evidence in an operation. Do we wipe out existing data? Can you select just specific sources? etc.

For re-running enrichment, I think it makes sense to put a button to trigger a re-run for a particular source on the metadata view modal, and also a button to trigger a first run for any enrichment pipelines that we happen to add. The backend "right now" is set up to run for a particular piece of evidence for a particular source. it's easy enough to extend this to all sources, but I was thinking that generally we don't want to re-run all workers, as the data is unchanged -- only the processor will change, and that's not likely to happen all at once.

For re-running enrichment on everything, that's probably something we want to avoid, but we can do it if we introduce a queuing mechanism for this work. Otherwise, I think we might overwhelm the pipelines.

Again, more of a pipeline task, which is why it's not addressed here.

> 3. Is a text blob for the metadata sufficient? How does that effect indexing and search? Do we need to do stemming or any sort of normalization or processing on the data before it goes in? If so, is that done by the api or the individual sources?

Is text blob sufficient? Well, it's the most flexible, besides a real blob. The `text` type can't be indexed, however, so that's not possible unless we switch to a full-text text field (Which can be indexed, but it's a bit different.).  However, I thought we were only planning on doing `metadata like '%whatever%'` (and `not like`), similar to how we work with descriptions. We can revisit that conversation. As for normalization, I don't know what we would do to normalize it. That might be something we decide on once we have a better idea of the output we're going to get from each of the planned sources. 

The one thought I did have is that it might make sense for the service to indicate what kind of data is being returned. Right now, I'm assuming we get back just raw text that doesn't have a particular form. That's not a great assumption, and it might make sense to have the pipeline service indicate what the content-disposition is, so that we can render the metadata in a more logical way (e.g. if it indicates `application/json` as the mimetype, we could render a formatted version of the data inside ace, as we do now for the codeblocks). I'm still under the impression that the data is always going to be textual. If you think otherwise, then I think we'll need to switch to true blob, and then content-disposition is probably a strict requirement.

JoelAtDeluxe:
Closing because the other PRs have eclipsed this PR.

PR #577, opened 2022-05-12
Metadata, Part 2 - Evidence Metadata CRUD support

JoelAtDeluxe:
This PR introduces the evidence metadata table, as well as the ability to interact with it. However, beyond providing that capability, no work is dedicated towards actually using the evidence metadata.

Note: This is part 2 of a series. This will be merged into metadata, part 1 (PR #576) without any changes. 


backend/migrations/20220512174013-evidence-metadata.sql
```diff
@@ -0,0 +1,20 @@
+-- +migrate Up
+CREATE TABLE evidence_metadata (
+    `id` INT AUTO_INCREMENT,
+    `evidence_id` INT NOT NULL,
+    `source` VARCHAR(255) NOT NULL,
+    `body` TEXT NOT NULL,
```
JoelAtDeluxe, line 6: Do we want to keep body, and introduce a full text search option, or do we want to change this to `varchar(65535)` and have a regular index?

JoelAtDeluxe: Probably not a `varchar` index. I can't find the explicit reference for this, but when you create an index on a varchar column, you specify how many characters you want to index. A side effect of this is that you can only take advantage of the index when the query is written like: `SELECT * FROM table WHERE varchar_column like 'text%'`. Most of our queries are instead `like '%text%'`, so if we assume metadata will want to be queried the same as description, then it stands to reason that our indexing won't be useful.

PR #580, opened 2022-05-16
Metadata, Part 4 - Support Service Workers

JoelAtDeluxe:
This PR introduces the concept of a service worker (but does not add any -- that will be a  future commit). This PR limits the scope of these workers to be more like microservices than lambdas. Specifically, these workers interact with AShirt in the following way:

1. An ashirt admin creates a headless user, with an API key and secret
2. The admin then spins up the worker, supplying the API credentials
3. The admin then goes to the service worker section, writes the config on how to contact the serviced, and registers the worker.

After, whenever any new evidence is added, ashirt will contact these services with a message effectively saying that "Evidence UUID, for operation slug, was added, with content type X", allowing the service to know if it should process the file, or ignore it. The service also has the option of either immediately processing the file, or deferring processing if it knows the processing will take some time. This all happens in a separate goroutine, allowing normal work to continue and letting the enhancement work as a background task.


backend/pipeline_readme.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Web Service
+
+Web services are any service that can be contacted via an HTTP request, and respond in kind. The details on exactly how the request is sent and what the body is are below. Note that this can include AWS Lambda services, though there is a more specific tool available if you want to use AWS Lambda.
+
+#### Web Configurations
+
+Web services, like all pipeline services, must define their configuration to AShirt when adding the service. The configuration provided below is mostly concerned with how to contact your service, and provides some ways for ashirt to minimally customize the message your service receives.
+
+##### Web, Version 1
```
jkennedyvz, line 48: Can we include an example config for the lambda service here also? (unless it is covered elsewhere)

jkennedyvz (approved):

jkennedyvz (approved):

PR #596, opened 2022-06-07
Upgrade AWS library to version 2

JoelAtDeluxe:
This PR changes the golang aws library to the v2 golang library, over the v1 library. It may be worthwhile to hold off on merging this PR until after the metadata PRs are merged, as the lambda introduction would also need to be updated.


JoelAtDeluxe:
This PR is now out of date. The addition of  lambdas means that we should spend some time trying to make those work. The interface changed slightly, but it might have some broader ramifications. I'm still going through the docs on this one, but my recommendation is to not merge at the moment.

jrozner:
Rebased on main and resolved conflicts. It looks like all the service worker code and lambdas were written against the new library rather than the old one. We can deploy this into staging and test it out before merging and do a pass over it to see if anything else needs to change.

jkennedyvz (approved):
LGTM

TylerNoblett (dismissed):
code looks good, works for me locally and on aws 🚀

jkennedyvz (approved):

PR #646, opened 2022-07-08
Js/better service tests

JoelAtDeluxe:
This PR creates some test helpers to properly clean up test resource and speed up tests for the services package in particular. The speed up is achieved by maintaining a single connection to the database and without re-creating the database on each test. Instead, tests will use the established connection and reset the database to a clean state, or if they need a truly clean state (as in the case with one test),, then there's another method that can be used to do what we did before.

Tests should be able 75% faster with this method, and we shouldn't run into issues where we utilize all database connections. There may be an even faster approach, but it requires a project-wide adjustment that is beyond the scope of this PR.

### DIFF
--- a/backend/services/evidence_test.go
+++ b/backend/services/evidence_test.go
@@ -22,145 +22,140 @@ import (
 type evidenceValidator func(*testing.T, FullEvidence, dtos.Evidence)
 
 func TestCreateEvidence(t *testing.T) {
-	db := initTest(t)
-	defer db.DB.Close()
-	HarryPotterSeedData.ApplyTo(t, db)
+	RunResettableDBTest(t, func(db *database.Connection, _ TestSeedData) {
+		memStore, _ := contentstore.NewMemStore()
+		normalUser := UserRon
+		ctx := contextForUser(normalUser, db)
 
-	memStore, _ := contentstore.NewMemStore()
-	normalUser := UserRon
-	ctx := contextForUser(normalUser, db)
+		op := OpChamberOfSecrets
+		imgContent := TinyImg
+		imgInput := services.CreateEvidenceInput{
+			OperationSlug: op.Slug,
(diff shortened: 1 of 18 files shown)


backend/services/evidence_test.go
```diff
@@ -373,39 +368,13 @@ func validateEvidenceSets(t *testing.T, dtoSet []dtos.Evidence, dbSet []FullEvid
...
 
 func toPtrTagList(in []dtos.Tag) []*dtos.Tag {
-	rtn := make([]*dtos.Tag, len(in))
-
-	for i := range in {
-		rtn[i] = &in[i]
-	}
-
-	return rtn
+	return helpers.Map(in, helpers.Ptr[dtos.Tag])
```
JoelAtDeluxe, line 371: Refactored these two methods because the generics we've created previously do exactly the same thing.

backend/services/evidence_test.go
```diff
@@ -175,173 +170,173 @@ func getAssociatedFindings(t *testing.T, db *database.Connection, evidenceID int
...
-	ctx := contextForUser(UserRon, db)
-	cs, _ := contentstore.NewMemStore()
+	RunResettableDBTest(t, func(db *database.Connection, _ TestSeedData) {
+		ctx := contextForUser(UserRon, db)
+		cs, _ := contentstore.NewMemStore()
 
-	masterOp := OpChamberOfSecrets
-	masterEvidence := cloneEvidence(EviFlyingCar) // cloning so we don't alter the underlying evidence for other tests
+		masterOp := OpChamberOfSecrets
+		masterEvidence := EviFlyingCar
```
JoelAtDeluxe, line 221: `cloneEvidence` really did nothing. Copying here is sufficient for our needs.

jkennedyvz (approved):

PR #696, opened 2022-09-15
Disable non-admin editing

TylerNoblett:
This PR disables UI that appears to let non-admin users
1.  Edit the permissions of other users (note that these actions are already disallowed on the backend, but the user can't tell that).
2. Add new users to the operation

### Old Version
(note that Ron is an admin in this screenshot, but it would look the same even if he wasn't)
Screen Shot 2022-09-15 at 10 23 51 AM

### New version
Screen Shot 2022-09-15 at 10 11 52 AM


jkennedyvz (approved):

JoelAtDeluxe (approved):
Looks good to me

PR #720, opened 2022-10-27
remove operation status attribute

TylerNoblett:
Operation status removed from frontend, backend, and db

NOTE: I based this branch from the `additional-card-data` branch to avoid a merge conflict (I'm assuming additional card data will get merged first (hopefully))


backend/migrations/20221027152757-remove-operation-status.sql
```diff
@@ -0,0 +1,7 @@
+-- +migrate Up
+ALTER TABLE operations
+  DROP COLUMN status;
+
+-- +migrate Down
+ALTER TABLE operations
```
JoelAtDeluxe, line 6: You probably need to add a default value here

backend/services/operations_test.go
```diff
@@ -213,16 +212,16 @@ func TestUpdateOperation(t *testing.T) {
 		input := services.UpdateOperationInput{
 			OperationSlug: masterOp.Slug,
 			Name:          "New Name",
-			Status:        models.OperationStatusComplete,
 		}
-		require.NotEqual(t, masterOp.Status, input.Status)
+		// TODO TN - look into this?
+		// require.NotEqual(t, masterOp.Status, input.Status)
```
JoelAtDeluxe, line 217: It looks like you can safely remove this, and the similar status checks here

TylerNoblett: oh whoops - forgot to remove those

backend/migrations/20221027152757-remove-operation-status.sql
TylerNoblett: I didn't see one in the initial table setup or in any migrations - I guess '0' would work, since that's the previous status code number for the 'planning' stage?

backend/services/operations_test.go
TylerNoblett: removed

backend/migrations/20221027152757-remove-operation-status.sql
TylerNoblett: changed to '0' as the default

frontend/src/vars.styl
```diff
@@ -21,8 +21,7 @@ $listen = #A66321
 $key-top = #222
 $key-bottom = #444
 
-$operation-status-colors = {
-  // Keys here are the backend representations of OperationStatus defined in backend/models
+$challenge-colors = {
   '0': #FFC940,
   '1': #3DCC91,
   '2': #FF7373,
```
JoelAtDeluxe, line 27: This makes sense for now, but we might want to change how this is represented. Also, we might not need all of the colors. The mapping was to help tie the status to the color, but since there's no status, it's probably better to call these something like $challenge-warning vs $challenge-colors[1]

JoelAtDeluxe (approved):
This looks good to me.

backend/migrations/20221027152757-remove-operation-status.sql
JoelAtDeluxe: Yeah, it probably didn't need a default value the first time, because no operations existed (assuming this was in the initial launch). 0 makes sense to me. We don't have room for nuance, so we're sort of stuck with limited options. 

An alternative if we didn't want to go with a default would be a 3 step approach: Add the column back as nullable, set the value for everything (again, not many options, so 0 makes sense here), then make the column non-nullable. Or variation on this.

frontend/src/vars.styl
TylerNoblett: changed

PR #743, opened 2022-12-05
Move favorites to (new) Operation Preferences

JoelAtDeluxe:
This PR revises how favorites are handled to allow for group access/rbac to work without having to consider a user's individual favorite preferences. This also allows for more operation-based preferences to be added in the future for individual users.

### DIFF
--- a/backend/services/operations.go
+++ b/backend/services/operations.go
@@ -423,9 +423,9 @@ func SetFavoriteOperation(ctx context.Context, db *database.Connection, i SetFav
 	}
 
 	_, err = db.Insert("user_operation_preferences", map[string]interface{}{
-		"user_id": middleware.UserID(ctx),
+		"user_id":      middleware.UserID(ctx),
 		"operation_id": operation.ID,
-		"is_favorite": i.IsFavorite,
+		"is_favorite":  i.IsFavorite,
 	}, "ON DUPLICATE KEY UPDATE is_favorite=VALUES(is_favorite)")
 
 	if err != nil {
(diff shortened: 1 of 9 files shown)


backend/services/operations_test.go
```diff
@@ -181,7 +181,7 @@ func TestListOperationsForAdmin(t *testing.T) {
 func TestSetFavoriteOperation(t *testing.T) {
 	RunResettableDBTest(t, func(db *database.Connection, _ TestSeedData) {
 		normalUser := UserRon
-		slug := "HPGoF"
+		slug := OpGobletOfFire.Slug
```
TylerNoblett, line 184: ooh good catch - I think that was my bad

backend/services/operations.go
```diff
@@ -417,11 +422,12 @@ func SetFavoriteOperation(ctx context.Context, db *database.Connection, i SetFav
...
-	err = db.Update(sq.Update("user_operation_permissions").
-		SetMap(map[string]interface{}{
-			"is_favorite": i.IsFavorite,
-		}).
-		Where(sq.Eq{"operation_id": operation.ID, "user_id": middleware.UserID(ctx)}))
+	_, err = db.Insert("user_operation_preferences", map[string]interface{}{
+		"user_id": middleware.UserID(ctx),
+		"operation_id": operation.ID,
+		"is_favorite": i.IsFavorite,
+	}, "ON DUPLICATE KEY UPDATE is_favorite=VALUES(is_favorite)")
```
TylerNoblett, line 429: Is that how you do an upsert? For my current PR I was trying to figure out if that was possible with the squirrel (it's not a supported method), but this might solve the problem for me!

TylerNoblett (approved):
Thanks for doing this Joel!

backend/services/operations.go
TylerNoblett: good to know!

jkennedyvz (approved):

PR #762, opened 2023-01-04
Tn/add groups

TylerNoblett:
I've added a user group feature that allows system admins to create groups of users, and then system admins or operation admins can assign those groups to an operation. If a user is assigned to an operation via a group AND individually, the greater of the permission levels of the two will be granted.

Here's a video showing the system admin view:
https://user-images.githubusercontent.com/22867407/210581893-9ec21a42-ca37-4699-980c-82e93ac4e3dd.mov

Here's a video showing how to add a user via a group
https://user-images.githubusercontent.com/22867407/210582049-c0a6ac7f-77f9-422c-b097-237333ebcfb0.mov

Here's a video showing that the greater permission level is granted 
https://user-images.githubusercontent.com/22867407/210582353-7758a9a4-82bd-4620-982e-14230b72cfbc.mov


backend/services/user_groups.go
```diff
@@ -0,0 +1,441 @@
...
+
+	paginatedResults := userGroupsDTO[prevLastIndex:currLastIndex]
+	paginatedData := &dtos.PaginationWrapper{
+		PageNumber: pagination.Page,
+		PageSize:   pageSize,
+		Content:    paginatedResults,
+		TotalCount: int64(groupLength),
+		TotalPages: int64(totalPages),
+	}
+	return paginatedData, nil
```
TylerNoblett, line 364: I can't rely on the normal pagination, because I'm receiving a table that lists each user and their affiliated group AND a row that just lists the table, in case there are no users assigned to the group, which returns way more rows than I need, so I had to write my own pagination logic

frontend/src/services/user_groups.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export async function listUserGroupsAdminView(i: ListObjectForAdminQuery): Promise<PaginationResult<UserGroupAdminView>> {
+  return await ds.adminListUserGroups(i)
+}
+
+export async function createUserGroup(i: {
+  name: string,
+  userSlugs: string[],
+}): Promise<void> {
+  let slug = i.name.toLowerCase().replace(/[^A-Za-z0-9]+/g, '-').replace(/^-|-$/g, '')
```
JoelAtDeluxe, line 20: Suppose my user group name was `$$bestgroup$$`, I think that would initially change this to `--bestgroup--`, then get fixed into just `-bestgroup-`? 
If that's the case, then...
It looks like we're trying to prevent user groups from starting/trailing -'s, so maybe we want to change the latter regex to `/^-+|-+$/g` (or collapse all multiple -'s into a single dash before doing the final stripping)

backend/database/seeding/helpers.go
```diff
@@ -245,6 +254,38 @@ func newUserOperationPreferences(user models.User, op models.Operation, isFavori
 	}
 }
 
+func newUserGroupGen(first int64) func(name string, deleted bool) models.UserGroup {
+	id := iotaLike(first)
+	return func(name string, deleted bool) models.UserGroup {
+		if deleted {
```
JoelAtDeluxe, line 260: nice

backend/database/seeding/helpers.go
```diff
@@ -245,6 +254,38 @@ func newUserOperationPreferences(user models.User, op models.Operation, isFavori
...
+				Slug:      name,
+				Name:      name,
+				CreatedAt: internalClock.Now(),
+			}
+		}
+
+	}
+}
+
+func newUserGroupMapping(userID int64, groupID int64) models.UserGroupMap {
```
JoelAtDeluxe, line 281: This is totally fine, but in the future you might want to use models.User and models.UserGroup. The benefit is very minor -- using UserHarry instead of UserHarry.id.

backend/server/middleware/authenticator.go
```diff
@@ -136,6 +136,17 @@ func buildPolicyForUser(ctx context.Context, db *database.Connection, userID int
...
+
+	var userGroupIds []int64
+	err = db.Select(&userGroupIds, sq.Select("group_id").
+		From("group_user_map").
+		Where(sq.Eq{"user_id": userID}))
+
+	var groupRoles []models.UserGroupOperationPermission
+	err = db.Select(&groupRoles, sq.Select("operation_id", "role").
+		From("user_group_operation_permissions").
+		Where(sq.Eq{"group_id": userGroupIds}))
```
JoelAtDeluxe, line 148: Two things:
1. you're not checking the errors here
2. This should maybe be done in a transaction. A transaction is not necessarily needed here, but it'll capture errors for you, which should allow you to ignore the errors and only check if the transaction itself succeeded. It'll also prevent weird "what if someone changes permissions between db statements"-type errors.

backend/services/helpers_test.go
```diff
@@ -0,0 +1,19 @@
+// Copyright 2022, Yahoo Inc.
```
JoelAtDeluxe, line 1: Year change? I guess I have the advantage of living in the present. I imagine this was correct when it was written. :)

backend/services/helpers_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/theparanoids/ashirt-server/backend/services"
+)
+
+func TestSanitizeSlug(t *testing.T) {
+	require.Equal(t, services.SanitizeSlug("?One?Two?Three?"), "one-two-three")
+	require.Equal(t, services.SanitizeSlug("Harry"), "harry")
+	require.Equal(t, services.SanitizeSlug("Harry Potter"), "harry-potter")
+	require.Equal(t, services.SanitizeSlug("fancy_name"), "fancy-name")
+	require.Equal(t, services.SanitizeSlug("Lots_Of-Fancy! Characters"), "lots-of-fancy-characters")
```
JoelAtDeluxe, line 18:
```suggestion
	require.Equal(t, services.SanitizeSlug("Lots_Of-Fancy! Characters"), "lots-of-fancy-characters")
	require.Equal(t, services.SanitizeSlug("$$prefixed_and_postfixed$$"), "prefixed-and-postfixed") 
```

That should test my "what if the user provides extra leading/ending unsupported characters" question (seems to work on my local branch!)

backend/services/helpers_test.go
TylerNoblett: Oh man - I would love to live in the present! haha yeah it got outdated so quickly 😂

frontend/src/services/user_groups.ts
TylerNoblett: that's a good point! This function is actually some copypasta from `createOperation` in `operations.ts` but slightly modified for usergroups. yeah I'm down for changing it (I guess in both places) to `/^-+|-+$/g`

backend/database/seeding/helpers.go
TylerNoblett: thanks! I can't take too much credit - most of this PR was just figuring out what other models (users, operations) were doing, and then taking it and modifying it slightly 😅

backend/database/seeding/helpers.go
TylerNoblett: oh yeah now I see that most of the other helpers that use IDs take the full object

frontend/src/services/user_groups.ts
TylerNoblett: changed

backend/database/seeding/helpers.go
TylerNoblett: changed

backend/services/helpers_test.go
TylerNoblett: nice!

backend/server/middleware/authenticator.go
TylerNoblett: oh whoops good catch!

TylerNoblett: changed

backend/services/operation_role.go
```diff
@@ -79,3 +85,62 @@ func SetUserOperationRole(ctx context.Context, db *database.Connection, i SetUse
...
+		err := db.Delete(sq.Delete("user_group_operation_permissions").Where(sq.Eq{"group_id": userGroupID, "operation_id": operation.ID}))
+
+		if err != nil {
+			return backend.WrapError("Cannot delete user group role", backend.DatabaseErr(err))
+		}
+		return nil
+	}
+
+	var permission models.UserGroupOperationPermission
+	err = db.Get(&permission, sq.Select("*").
```
JoelAtDeluxe, line 118: This should be done in a transaction

backend/services/user_groups.go
```diff
@@ -0,0 +1,441 @@
...
+	Role policy.OperationRole `db:"role"`
+}
+
+type ListUserGroupsInput struct {
+	Query          string
+	IncludeDeleted bool
+	OperationSlug  string
+}
+
+func (cugi ModifyUserGroupInput) validateUserGroupInput() error {
```
JoelAtDeluxe, line 62: the `cugi` should maybe be renamed to `mugi` (or maybe just `i` or something similar?).

backend/services/user_groups.go
```diff
@@ -0,0 +1,441 @@
...
+		if len(userGroupMap) == 0 {
+			_, err = db.Insert("group_user_map", map[string]interface{}{
+				"user_id":  userID,
+				"group_id": groupID,
+			})
+			if err != nil {
+				return backend.WrapError("Unable to connect user to group", backend.DatabaseErr(err))
+			}
+		}
+	}
```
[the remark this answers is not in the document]
JoelAtDeluxe: oh, I see, it _is_ in a transaction. Great, but still, maybe there's a more optimal solution

backend/services/user_groups.go
```diff
@@ -0,0 +1,441 @@
...
+	})
+
+	if err != nil {
+		return nil, backend.WrapError("Error creating user group", backend.BadInputErr(err, "A user group with this slug already exists; please choose another name"))
+	}
+	return &dtos.UserGroup{
+		Slug: i.Slug,
+		Name: i.Name,
+	}, nil
+}
```
JoelAtDeluxe, line 138: I like this function. it looks super clean.

backend/services/user_groups.go
```diff
@@ -0,0 +1,441 @@
...
+
+	err = db.WithTx(context.Background(), func(tx *database.Transactable) {
+		if i.Name != "" {
+			tx.Update(sq.Update("user_groups").Set("name", i.Name).Where(sq.Eq{"id": userGroup.ID}))
+		}
+		if len(i.UsersToRemove) > 0 {
+			for _, userSlug := range i.UsersToRemove {
+				var userID int64
+				tx.Get(&userID, sq.Select("id").From("users").Where(sq.Eq{"slug": userSlug}))
+				tx.Delete(sq.Delete("group_user_map").Where(sq.Eq{"user_id": userID, "group_id": userGroup.ID}))
```
JoelAtDeluxe, line 162: I need to double check, but you might be able to combine these queries. The sql might look like:
```sql
DELETE FROM group_user_map JOIN users on group_user_map.user_id = users.id WHERE users.slug=?
```

frontend/src/components/checkbox_complex/index.tsx
```diff
@@ -0,0 +1,27 @@
+// Copyright 2020, Yahoo Inc.
```
JoelAtDeluxe, line 1:
```suggestion
// Copyright 2023, Yahoo Inc.
```

backend/services/user_groups.go
TylerNoblett: ooh good catch - yeah that was copy pasta

backend/services/user_groups.go
TylerNoblett: Thanks!

backend/services/user_groups.go
TylerNoblett: oh of course! Yeah I think that looks correct; I'll try it out

TylerNoblett: changed

backend/services/operation_role.go
TylerNoblett: changed

backend/services/operation_role.go
```diff
@@ -79,3 +85,63 @@ func SetUserOperationRole(ctx context.Context, db *database.Connection, i SetUse
...
+	}
+
+	if len(permissions) > 0 && permissions[0].Role != i.Role {
+		err = db.Update(sq.Update("user_group_operation_permissions").
+			Set("role", i.Role).
+			Where(sq.Eq{"group_id": userGroupID, "operation_id": operation.ID}))
+
+		if err != nil {
+			return backend.WrapError("Unable to alter user role", backend.DatabaseErr(err))
+		}
```
TylerNoblett, line 144: I moved these operations into a transaction liked you asked (which I think was the right call!). Before. I was using a `tx.Get(&permission,...` to get the permissions, and if there was an error (because there was no record), then I'd do the insertion `db.Insert("user_group_operation_permissions", ...`.

However, now that I'm in a transaction I can't rely on the same logic, because if my `Get` doesn't return anything the transaction fails, so I changed `var permission models.UserGroupOperationPermission` to `var permissions []models.UserGroupOperationPermission`, and am checking the length of the slice. This leads to some clunkly logic like `if len(permissions) > 0 && permissions[0].Role != i.Role` instead of `if permission.Role != i.Role {`. 

I've tried to figure out if there's a helper I could create that returns nil if the `Get` is empty, but it seems I would need to modify some code in the Sqlx library, which seems doesn't seem like an option.

Is there a different way you would do this, or does that seem fine?

backend/services/user_groups.go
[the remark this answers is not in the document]
TylerNoblett: ah sorry let me change that to be `tx` for clarity

Yeah let me see if I can figure out a way to roll it into one!

backend/services/user_groups.go
```diff
@@ -76,33 +76,31 @@ func (i CreateUserGroupInput) validateUserGroupInput() error {
...
 }
 
-func AddUsersToGroup(tx database.ConnectionProxy, userSlugs []string, groupID int64) error {
-	for _, userSlug := range userSlugs {
-		var userID int64
-		err := tx.Get(&userID, sq.Select("id").From("users").Where(sq.Eq{"slug": userSlug}))
-		if err != nil {
-			return backend.WrapError("Unable to get user id from slug", backend.BadInputErr(err, fmt.Sprintf(`No user with slug %s was found`, userSlug)))
+func AddUsersToGroup(tx *database.Transactable, userSlugs []string, groupID int64) error {
+	if len(userSlugs) > 0 {
```
TylerNoblett, line 80: @JoelAtDeluxe Initially I improved this function by iterating through each userSlug and performing an upsert, but here's my attempt to reduce it to one query. It looks a little janky, but (in theory) should be much more performant that inserting each userSlug individually. Any way you'd improve this?

backend/services/user_groups.go
```diff
@@ -79,9 +79,13 @@ func (i CreateUserGroupInput) validateUserGroupInput() error {
 func AddUsersToGroup(tx *database.Transactable, userSlugs []string, groupID int64) error {
```
TylerNoblett, line 79: @JoelAtDeluxe Initially I improved this function by iterating through each userSlug and performing an upsert, but here's my attempt to reduce it to one query. It looks a little janky, but (in theory) should be much more performant that inserting each userSlug individually. Any way you'd improve this?

backend/services/user_groups.go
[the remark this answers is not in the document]
TylerNoblett: Initially changed to upsert each userSlug, then figured out a way to [combine them all into one query](https://github.com/theparanoids/ashirt-server/pull/762#discussion_r1093704784)

frontend/src/pages/admin_modals/index.tsx
```diff
@@ -169,6 +174,130 @@ export const AddUserModal = (props: {
   )
 }
 
+export const AddUserGroupModal = (props: {
+  onRequestClose: () => void,
+}) => {
+  const [isCompleted, setIsCompleted] = React.useState<boolean>(false)
+  const [includedUsers, setIncludedUsers] = React.useState(() => new Set());
```
JoelAtDeluxe, line 181: I think you should be able to modify this to remove the `as Set<string>` lines:

```suggestion
  const [includedUsers, setIncludedUsers] = React.useState<Set<string>>(() => new Set());
```
It's about the same as what you have though.

backend/services/operation_role.go
JoelAtDeluxe: Short version: this is _probably_ the best bet. there are a couple of other routes:
1. Add a clearError function to transactable -- this is handy, but potentially dangerous.
2. Add a special Get function that calls Sqlx.Get under the hood, but checks explicitly for zero row or more than one row, and alerts the caller appropriately -- Still an error, but, one you can check against, and it would explicitly NOT set the transactable error

And I think there are some variations of the above. I'd say it's worth adding that logic if you think you'll need it later, or when you do need it later. For now, this looks okay. You might consider moving the lower if condition (`if len(permissions) > 0` inside the transaction though for additional safety. It would also mean that you don't need to check the length, because it'll be as an `else if` block.

TylerNoblett: > You might consider moving the lower if condition (if len(permissions) > 0 inside the transaction though for additional safety. It would also mean that you don't need to check the length, because it'll be as an else if block.

oh great point! Thanks for the suggestions - yeah I'll leave it that way for now if since that sounds like a decent route

backend/services/user_groups.go
```diff
@@ -0,0 +1,421 @@
...
+
+	err = db.WithTx(context.Background(), func(tx *database.Transactable) {
+		if i.Name != "" {
+			tx.Update(sq.Update("user_groups").Set("name", i.Name).Where(sq.Eq{"id": userGroup.ID}))
+		}
+		if len(i.UsersToRemove) > 0 {
+			for _, userSlug := range i.UsersToRemove {
+				tx.Exec(sq.Expr("DELETE gm FROM group_user_map gm JOIN users u on gm.user_id = u.id WHERE u.slug=?;", userSlug))
+			}
+		}
```
JoelAtDeluxe, line 158: I think you might be able to replace the loop with:

```go
tx.Exec(sq.Expr("DELETE gm FROM group_user_map gm JOIN users  u  ON gm.user_id = u.id WHERE u.slug in(?)", i.UsersToRemove))
```
Squirrel seems to support lists as parameters, so it might be possible. I didn't check though.

Unfortunately, it looks like [This  PR](https://github.com/Masterminds/squirrel/pull/295) needs to be merged before you can use the delete query builder.

backend/services/user_groups.go
```diff
@@ -0,0 +1,421 @@
...
+	if len(userSlugs) > 0 {
+		questionMarks := "("
+
+		interfaceSlice := make([]interface{}, len(userSlugs)+1)
+		for i, v := range userSlugs {
+			questionMarks += "?, "
+			interfaceSlice[i] = v
+		}
+
+		questionMarks = strings.TrimSuffix(questionMarks, ", ")
```
JoelAtDeluxe, line 89: I think you can use [strings.Join](https://pkg.go.dev/strings#Join) for this, but I think there are bigger improvements here.

backend/services/user_groups.go
JoelAtDeluxe: Ah, hm, actually this might not work like I expect --  you might need to use  Sq.Eq, which would _probably_ mean that you'd have to use a deleteBuilder, which, per the PR above, won't work. :( You might be able to do something like you did in addUsers though -- joining the slugs and appending a dynamic number of elements.

backend/services/user_groups.go
```diff
@@ -0,0 +1,421 @@
...
+	}
+	if i.Name == "" {
+		return backend.MissingValueErr("Name")
+	}
+	return nil
+}
+
+func AddUsersToGroup(tx *database.Transactable, userSlugs []string, groupID int64) error {
+	if len(userSlugs) > 0 {
+		questionMarks := "("
```
JoelAtDeluxe, line 81: This is a style thing, but Go tends to prefer early exits, so you'd normally see:
```go
if len(userSlugs) == 0 {
  return nil
}
```
Doesn't really matter though.

backend/services/user_groups.go
JoelAtDeluxe: I think there is a better way, but we  haven't leveraged this in the code before. I'll post a suggestion  with some details.

backend/services/user_groups.go
```diff
@@ -0,0 +1,421 @@
...
+		err := tx.Exec(sq.Expr(sqlStatement, interfaceSlice...))
+
+		if err != nil {
+			return backend.WrapError("Unable to add users to group", backend.DatabaseErr(err))
+		}
+		return nil
+	}
+
+	return nil
+}
```
JoelAtDeluxe, line 107:
```suggestion
func AddUsersToGroup(tx *database.Transactable, userSlugs []string, groupID int64) error {
	if len(userSlugs) == 0 {
		return nil
	}

	peopleToAdd := sq.Select("id", strconv.FormatInt(groupID, 10)).
		From("users").
		Where(sq.Eq{"slug": userSlugs})

	insertQuery := sq.
		Insert("group_user_map").
		Options("Ignore").
		Columns("user_id", "group_id").Select(peopleToAdd)

	err := tx.Exec(insertQuery)

	if err != nil {
		return backend.WrapError("Unable to add users to group", backend.DatabaseErr(err))
	}
	return nil
}
```

Here's what's going on:

We can't use the  normal Insert or BatchInsert functions because those are more designed for normal insert cases, which this isn't. Instead, we're going to use what those use: the Squirrel InsertBuilder and tx.Exec.

Let's start on the insertQuery line: We're going to start a normal insert, but we're going to tack on the ignore option, which should get us `INSERT IGNORE INTO`.We'll then specify the columns as usual, but instead of specifying the values, we'll defer to a select statement.
Back on the peopleToAdd line, we're going to select the user's id for a given slug, but we're _also_ going to select a fixed value so that we end up with a series of (user_id/user_group_id) pairs.

Adding those two things together, we get a single sql query (unexecuted) that should get us what we want. tx.Exec makes it official, and everything else is basically the same.

Really, this is just doing what you hand coded. The only "significant" change is that I'm selecting a constant value, rather than searching the user_group table for an id I already know. I'd imagine that behind the scenes sql will somewhat optimize that query, so it wouldn't  have been a slow solution, just a strange one.

backend/services/user_groups.go
```diff
@@ -0,0 +1,441 @@
...
+	err := db.Select(&slugMap, unionSelect)
+
+	if err != nil {
+		return nil, backend.WrapError("unable to get map of user IDs to group IDs from database", backend.DatabaseErr(err))
+	}
+
+	return slugMap, nil
+}
+
+func SortUsersInToGroups(slugMap SlugMap, pagination Pagination) (*dtos.PaginationWrapper, error) {
```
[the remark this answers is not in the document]
JoelAtDeluxe: You could probably opt to not do pagination at all here, if that would make things easier, but I'll go through it. For reference, I added this when I started getting worried that there might be places where we want to add pagination, but mostly we don't use  it, and we don't even use it for listing all of the evidence in an operation, which is really a lot worse than getting a list of users.

I do think, generally, that we should paginate, but I'm not sure it's adding all that much value vs getting everything, and implementing some basic paging on the frontend only.

backend/services/user_groups.go
JoelAtDeluxe: uuuuuh, yeah. Maybe let's drop pagination? Or let's just stick with what you have here? (I have a slight preference for the former --  if we have to maintain this later, then I think we'd be better off with a simple, unoptimized solution).

backend/services/user_groups.go
```diff
@@ -0,0 +1,421 @@
...
+// no pagination, because this is used for the search bar
+func ListUserGroups(ctx context.Context, db *database.Connection, i ListUserGroupsInput) ([]*dtos.UserGroupAdminView, error) {
+	operation, err := lookupOperation(db, i.OperationSlug)
+	if err := policyRequireWithAdminBypass(ctx, policy.CanListUserGroupsOfOperation{OperationID: operation.ID}); err != nil {
+		return nil, backend.WrapError("Unwilling to list usergroups", backend.UnauthorizedReadErr(err))
+	}
+
+	if strings.ContainsAny(i.Query, "%_") || strings.TrimFunc(i.Query, unicode.IsSpace) == "" {
+		return []*dtos.UserGroupAdminView{}, nil
+	}
```
JoelAtDeluxe, line 395: Do we actually need this? To me, it looks like we're effectively saying that a group with an underscore isn't findable. I don't know if there are other rules that preclude that possibility, but if we _can_ support it, it seems like we should.

One suggestion: maybe instead of removing % and _, you could instead use `strings.ReplaceAll(i.Query, "%", "\\%)` (ditto with underscore). I think that'll properly escape them.

In any event, this is more fine tuning than an actual issue as far as I'm concerned.

JoelAtDeluxe (dismissed):
Sorry for the long, long time it took to get through this.

Overall, this looks fine. I think we can improve the add users to group query, and the deleting users from group query (into, hopefully, a single query, or at worst a group of queries (select from users, then delete where in) ). Personally, I think it's also worth dropping the pagination support on the backend for groups, and maybe just doing front-end only pagination.

With those two queries (and optionally the pagination) fixed, I think we'd be in a good place to merge.

backend/services/user_groups.go
TylerNoblett: yeah, I included this because there's a similar check for users (users.go 324):
```go
func ListUsers(ctx context.Context, db *database.Connection, i ListUsersInput) ([]*dtos.User, error) {
	if strings.ContainsAny(i.Query, "%_") || strings.TrimFunc(i.Query, unicode.IsSpace) == "" {
		return []*dtos.User{}, nil
	}
```
but I wasn't really sure if it was a security issue or what, so I'm down for removing it if it seems good to you as well

backend/services/user_groups.go
TylerNoblett: noted!

backend/services/user_groups.go
TylerNoblett: oh nice - yeah this is much cleaner

backend/services/user_groups.go
JoelAtDeluxe: I don't _think_ it's a security issue. those two characters are just placeholders in like queries -- % for  effectively `.*` and _ for effectively `.` (in regex speak). They could mess up the query -- which is why I was proposing that we escape them instead. I think that % and _ are unlikely username characters, but they might be more likely group name characters...

I'm now inclined to leave it as is just because we're already doing this somewhere else. Maybe a separate effort is done to see if we can remove those characters (or if we can even create groups with those characters)

TylerNoblett: sounds good!

backend/services/user_groups.go
TylerNoblett: modified it to use the ?'s like I was previously doing in addUsers!

jkennedyvz (dismissed):

TylerNoblett:
> Sorry for the long, long time it took to get through this.
> 
> Overall, this looks fine. I think we can improve the add users to group query, and the deleting users from group query (into, hopefully, a single query, or at worst a group of queries (select from users, then delete where in) ). Personally, I think it's also worth dropping the pagination support on the backend for groups, and maybe just doing front-end only pagination.
> 
> With those two queries (and optionally the pagination) fixed, I think we'd be in a good place to merge.

Thanks for reviewing it Joel - no worries; it was pretty long 😬 I've modified both queries and am going to take a look at pagination right now!

jkennedyvz (approved):

PR #777, opened 2023-02-07
update columns for docker prod works

TylerNoblett:
This PR corrects a few outdated columns - the effect of this is that the docker-compose-prod file was erroring out 


TylerNoblett:
> Is this seed data not used with the dev environment? Why was it loading before?

good question! the dev env uses different seed data `hp_seed_data.go`

Issue #781, opened 2023-02-24
Webauthn Login Failure


jkennedyvz:
I used ashirt-deployments to bring up the tip of main `sha-1bf765e`
When registering a new user with only webauthn the process seems to complete, but initial login fails with a 403 to the client. The username in this case was lower-alpha only. The uses appear in the database/admin settings, but the user cannot login. If I impersonate the user with a recovery link and re-add the same key the user can login. 

timestamp=2023-02-24T08:33:28.603963926Z ctx=fb404bcf-e62a-45c7-9918-06a3fd8e4ab9 msg="Error handling request" error="userHandle and User ID do not match" rootError="userHandle and User ID do not match" status=403 url=/auth/webauthn/login/finish

jrozner:
Fixed by #800

PR #800, opened 2023-03-30
Tn/fix authn bug

TylerNoblett:
Within login.go (of the webauthn library), there was a check that was failing:
```go
userHandle := parsedResponse.Response.UserHandle
	if len(userHandle) > 0 {
		a := user.WebAuthnID()
		if !bytes.Equal(userHandle, user.WebAuthnID()) {
			return nil, protocol.ErrBadRequest.WithDetails("userHandle and User ID do not match")
		}
	}
```
When I looked into it more, userHandle is a uuid generated when a new webauthn user is created that was saved in the session data (but wasn't saved in the database), and the WebAuthnID was the ID in the database (ie `26`). I took me a while to figure out exactly what was going, but saving the uuid and getting  `user.WebAuthnID()` to return that solved the issue.

I'm uncertain though of what caused this issue, as the library version we were using didn't change, and when I rolled the commit history back to when webauthn was first added, I still got the same error that [John encountered](https://github.com/theparanoids/ashirt-server/issues/781) 🤷🏼‍♂️


jkennedyvz (dismissed):

jrozner:
Changes here look good. I'm able to register and login. Did discover that `AUTH_WEBAUTHN_AUTHENTICATOR_RESIDENT_KEY` does not seem to be respected but we can create a new issue for that. We also probably want to rework the UI a bit and make some changes to webauthn in the future, specifically around supporting level 3 however the go library doesn't support it yet.

jrozner (approved):

PR #801, opened 2023-03-30
Tn/use new authn lib

TylerNoblett:
The [authn library](https://github.com/duo-labs/webauthn) has been archived, so I changed a few things to use the recommended fork


frontend/src/authschemes/webauthn/helpers.ts
```diff
@@ -22,6 +22,25 @@ export const toKeycode = (c: string) => c.charCodeAt(0)
 
 export const toByteArray = (s: string) => Uint8Array.from(atob(s), toKeycode)
 
+const base64UrlToBase64 = (input: string) => {
```
TylerNoblett, line 25: the only substantial change is the the 'challenge' string is now being sent in base64url (which is just base 64 but url safe; `–` is used instead of `+` and `_` is used instead of `/`) and there are some padding differences

jkennedyvz (approved):

PR #828, opened 2023-05-07
Remove clockwork

jrozner:
Now that clockwork has switched over to using the current time as it's fake clock there is no benefit unless we were to set a constant time. The way our tests are written it doesn't seem like we actually need it anyway.

- Please review our [contributing guidelines](https://github.com/theparanoids/ashirt-server/blob/master/Contributing.md)
- Please review our [Code of Conduct](https://github.com/theparanoids/ashirt-server/blob/master/Code-of-Conduct.md)

### DIFF
--- a/backend/database/seeding/helpers.go
+++ b/backend/database/seeding/helpers.go
@@ -11,28 +11,12 @@ import (
 	"time"
 
 	sq "github.com/Masterminds/squirrel"
-	"github.com/jonboulle/clockwork"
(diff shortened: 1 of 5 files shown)


JoelAtDeluxe (approved):
Tests seem to pass, and the logic doesn't appear  to be related to strict times. This is probably safe, and easily fixable later if we run into fuzzing-related issues.

Issue #734, opened 2022-11-27
Export Operation to Zip File


vysecurity:
# Feature Request

## Name

Ability to export an Operation to a Zip file, including all images, texts, screen recordings, terminal recordings.

## Brief Description

Often we want to backup operational data. It would be good to be able to download all the uploaded evidence into a Zip for archival purpose. Furthermore, timestamp prepend all of the file names.

It would also be good to be able to introduce screen recordings into the evidence.

JoelAtDeluxe:
The code is partially set up to do this, however there are a number of questions about how this would work in practice. This was an idea that was kicked around at one point, but was dropped for some reason.

Some questions that would eventually need to be answered:

* Is this export something that can be re-imported into ashirt-server?
* Who is allowed to perform an export?
* What happens to the operation on ashirt-server after someone archives it?
* what happens to the files on the file store/ s3?
* What is the utility of the export? Meaning, is this _just_ a backup, or is there some other desired utility?
* would we need/want to encrypt the zip/export format?
* is this a feature we would want to turn off for some reason?

jkennedyvz:
> * Is this export something that can be re-imported into ashirt-server?
I believe the idea here is not to migrate an op to a new instance, but to share raw data with a 3rd party. 

> * Who is allowed to perform an export?
We probably want to limit this to operation or system admins. 

> * What happens to the operation on ashirt-server after someone archives it?
@vysecurity feel free to answer this one.  We already have the capability to delete an operation and all of its evidence, so I don't feel any automated action is necessary.

> * what happens to the files on the file store/ s3?
> * What is the utility of the export? Meaning, is this _just_ a backup, or is there some other desired utility?
see above

> * would we need/want to encrypt the zip/export format?
Considering the sensitivity of the data in ashirt we should definitely encrypt any bulk export. We should also store logs of who has created a backup somewhere in the operation settings or detailed information. 

> * is this a feature we would want to turn off for some reason?
I could see teams wanting to turn this off entirely depending on culture, but we don't have that requirement.

jrozner:
The PR and branch that existed was #6. It doesn't look like we ever created an issue that defined the feature.

TylerNoblett:
>  but it was a lot of code to maintain and a lot of complexity that we felt we weren't going to get significant value from or was worth the effort to support.

Now that we are interested in adding back zip file support, do you guys still feel the same regarding this statement above?
I figure I can either
1) resurrect PRs 2,3 and 6
[...]

PR #831, opened 2023-05-08
Tn/add export functionality

TylerNoblett:
Adds an export feature per the [following features](https://github.com/theparanoids/ashirt-server/issues/734#issuecomment-1532289719)

A few things to note:
- a user can export the evidence if the global flag for exporting is enabled AND if they are an admin (system, operation, or part of a group with admin privileges).
- Re: denormlization, I removed the tag ID and the evidence UUID, but let me know if you'd like me to include the evidence UUID and I'll add that back.
- I am saving the images as JPEG and all the other media as txt, but I'm not sure if the terminal recordings are supposed to be another file type; let me know what they should be saved as!

https://user-images.githubusercontent.com/22867407/236897195-5e98f6c6-f577-4eb0-900c-9c1df3d6cad3.mov


docker-compose.yml
```diff
@@ -80,6 +80,7 @@ services:
     environment:
       - NODE_ENV=development
       - WEB_BACKEND_ORIGIN=http://backend:3000
+      - ENABLE_EVIDENCE_EXPORT=false
```
TylerNoblett, line 83: Global flag

frontend/src/pages/operation_show/evidence_list/index.tsx
```diff
@@ -61,14 +72,62 @@ export default () => {
     slug: operationSlug
   })
 
+  const createJsonEvidence = (evidence: Evidence[]) => {
+    const evidenceCopy = _.cloneDeep(evidence)
+    evidenceCopy.forEach(e => {
+      e.tags.forEach(t => {
+        delete (t as DenormalizedTag).id
+        })
+      delete (e as DenormalizedEvidence).uuid
```
TylerNoblett, line 81: removes the 
- tag ID 
- UUID  
let me know if you DO want UUID included

frontend/src/pages/operation_show/layout/toolbar/index.tsx
```diff
@@ -45,6 +47,8 @@ export const Toolbar = (props: {
       ]), [props.operationSlug]
     ))
 
+  const userCanExport = process.env.ENABLE_EVIDENCE_EXPORT && props.userCanExportData
```
TylerNoblett, line 50: allows user to export if exporting is enabled AND if they are an admin (system, operation, or part of a group with admin privileges)

frontend/webpack.config.js
```diff
@@ -60,6 +61,11 @@ module.exports = (env, argv) => ({
       filename: 'main-[contenthash].css',
       chunkFilename: '[chunkhash].css',
     }),
+    new webpack.DefinePlugin({
+      'process.env': {
+        'ENABLE_EVIDENCE_EXPORT': process.env.ENABLE_EVIDENCE_EXPORT == 'true',
```
TylerNoblett, line 66: one thought that I just had is that I could include this global flag on the backend and incorporate it into `UserCanExportData`, since most of the global flags are on the backend/makes more intuitive sense. I'll do that tomorrow morning

jrozner:
Those changes look good. If C/C++ are the only case where the extension isn't clear that seems fine to just use `.c`. If there are many cases where we don't know the exact extension, let's just not add an extension and simplify it.

One other change, can we remove the extra tag information (eg. color) and just make it an array of the tags themselves (eg. ["tag1", "tag2", "tag3"])?

jrozner:
Also, can we change the name from `uuid` to `filename`?

TylerNoblett:
Got it! I've
- removed the tag color
- changed uuid to filename
- I also added some extra logic to change the file extension to `.cpp` if it's a c/c++ file and the source file is listed with an extension of c++

jrozner:
Sorry, I think there was a misunderstanding with the tags. Can you just extract all the names from the tags and make it an array of strings (of the names) rather than an array of objects.

TylerNoblett:
> I think there was a misunderstanding with the tags. Can you just extract all the names from the tags and make it an array of strings (of the names) rather than an array of objects.

oh gotcha - I just reread your original message

backend/services/operations.go
```diff
@@ -283,6 +284,18 @@ func ReadOperation(ctx context.Context, db *database.Connection, operationSlug s
...
 
+	var userIsAdmin bool
+	if err := policyRequireWithAdminBypass(ctx, policy.CanExportOperationData{OperationID: operation.ID}); err == nil {
+		userIsAdmin = true
+	} else {
+		userIsAdmin = middleware.IsAdmin(ctx)
+	}
+
+	var userCanExportData bool
+	if os.Getenv("ENABLE_EVIDENCE_EXPORT") == "true" && userIsAdmin {
```
JoelAtDeluxe, line 295: This should probably go in webconfig.go and have a function to retrieve the parsed value.
Also, you could do this:

`userCanExportData := os.Getenv("ENABLE_EVIDENCE_EXPORT") == "true"  && userIsAdmin`

backend/dtos/dtos.go
```diff
@@ -69,6 +69,7 @@ type Operation struct {
 	TopContribs       []TopContrib  `json:"topContribs"`
 	EvidenceCount     EvidenceCount `json:"evidenceCount,omitempty"`
 	UserCanViewGroups *bool         `json:"userCanViewGroups,omitempty"`
+	UserCanExportData *bool         `json:"userCanExportData,omitempty"`
```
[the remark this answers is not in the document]
TylerNoblett: yeah I just copied what was there (and figured there was a good reason for it), but normal bool is fine with me 😄

backend/services/operations.go
TylerNoblett: ah did not know that existed - thanks for the heads up!

frontend/src/pages/operation_show/evidence_list/index.tsx
```diff
@@ -61,14 +73,96 @@ export default () => {
...
+    const evidenceCopy = _.cloneDeep(evidence)
+    const [media, modEvidenceCopy] = await getMedia(evidenceCopy)
+
+    zip.file("evidence.json", JSON.stringify(modEvidenceCopy))
+
+    if (media){
+      media.forEach((mb) => {
+        let filename;
+        const fileExtension = mb.sourceFilename?.slice(mb.sourceFilename?.lastIndexOf(".") + 1)
+        console.log("fileExtension", fileExtension)
```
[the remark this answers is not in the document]
TylerNoblett: ah no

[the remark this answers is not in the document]
TylerNoblett: addressed in [832](https://github.com/theparanoids/ashirt-server/compare/tn/joels-feedback-on-832)

PR #836, opened 2023-05-15
fix webauthn not working as second login

TylerNoblett:
When linking a webauthn account to an account created with the ashirt login, a user was unable to login using the webauthn account; this PR fixes that

### DIFF
--- a/backend/authschemes/webauthn/webauthn.go
+++ b/backend/authschemes/webauthn/webauthn.go
@@ -215,8 +215,13 @@ func (a WebAuthn) BindRoutes(r *mux.Router, bridge authschemes.AShirtAuthBridge)
 		if err != nil {
 			return nil, backend.WrapError("Unable to validate registration data", err)
 		}
+
+		rawSessionData := bridge.ReadAuthSchemeSession(r)
+		sessionData, _ := rawSessionData.(*webAuthNSessionData)
+
 		return nil, bridge.CreateNewAuthForUser(authschemes.UserAuthData{
 			UserID:   byteSliceToI64(data.UserData.UserID),
+			AuthnID:  sessionData.UserData.AuthnID,
 			Username: data.UserData.UserName,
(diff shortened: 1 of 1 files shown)


jrozner:
I wasn't able to reproduce the original bug this fixes. I logged into `Albus` using the username/password, added a webauthn credential, logged out, then logged in using webauthn. Is there another step to reproduce?

JoelAtDeluxe:
I think I might have fixed this in this  PR: https://github.com/theparanoids/ashirt-server/pull/829

TylerNoblett:
> I think I might have fixed this in this PR: #829

Oh interesting - I'm getting the error on main right now (with 829 merged in), but this does raise an interesting thing I've wondered (though it doesn't make a lot of sense); I wonder if different keys work somewhat differently with the webauthn library/spec. 

I say this because when I looked into the issue that John and I got, the webauthn code for ashirt hadn't changed, and we hadn't upgraded to a newer version of the go webauthn library, but both of us were getting the 403 forbidden error, AND when I went back to the commit in september where webauthn had been added, I couldn't get it to work (but obviously it had been working for at least Joel and I would assume either Joe or John). 

Now that we have a second instance where one version of the code worked some some people, but not another, it makes me wonder if the hardware key itself has anything to do with it (I'm using the touch ID sensor on my macbook fwiw). but IDK why that would be the case; what do you guys think?

JoelAtDeluxe:
I've been using a yubikey (I think a series 5). I've used it both with a fiingerprint biometric, and no biometric. I'm unclear why the different systems would behave differently, but I suppose it's something to look into.

TylerNoblett:
> I've been using a yubikey (I think a series 5). I've used it both with a fiingerprint biometric, and no biometric. I'm unclear why the different systems would behave differently, but I suppose it's something to look into.

yeah I'll take a look into it after I figure out some of this gorilla/mux stuff!

jkennedyvz (approved):

PR #842, opened 2023-05-24
adds api endpoint for additional evidence details

audibleblink:
While creating a service worker, I found myself needing more details about the evidence that was being created, like tags and the operator. 

### DIFF
--- a/backend/server/api.go
+++ b/backend/server/api.go
@@ -151,10 +151,16 @@ func bindAPIRoutes(r chi.Router, db *database.Connection, contentStore contentst
 
 	route(r, "GET", "/operations/{operation_slug}/evidence", jsonHandler(func(r *http.Request) (interface{}, error) {
 		dr := dissectJSONRequest(r)
-		tlq, _ := helpers.ParseTimelineQuery(dr.FromQuery("q").AsString())
+		tlq, err := helpers.ParseTimelineQuery(dr.FromQuery("query").AsString())
+		if err != nil {
+			return nil, err
+		}
 		i := services.ListEvidenceForOperationInput{
 			OperationSlug: dr.FromURL("operation_slug").Required().AsString(),
 			Filters:       tlq,
+		}
+		if dr.Error != nil {
+			return nil, dr.Error
 		}
 		return services.ListEvidenceForOperation(r.Context(), db, i)
 	}))
--- a/backend/server/api.go
+++ b/backend/server/api.go
@@ -15,6 +15,7 @@ import (
 	"github.com/theparanoids/ashirt-server/backend/contentstore"
 	"github.com/theparanoids/ashirt-server/backend/database"
 	"github.com/theparanoids/ashirt-server/backend/dtos"
+	"github.com/theparanoids/ashirt-server/backend/helpers"
 	"github.com/theparanoids/ashirt-server/backend/logging"
 	"github.com/theparanoids/ashirt-server/backend/server/middleware"
 	"github.com/theparanoids/ashirt-server/backend/services"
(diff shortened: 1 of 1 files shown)


backend/server/api.go
```diff
@@ -152,6 +153,16 @@ func bindAPIRoutes(r *mux.Router, db *database.Connection, contentStore contents
 		return services.ListTagsForOperation(r.Context(), db, i)
 	}))
 
+	route(r, "GET", "/api/operations/{operation_slug}/evidence", jsonHandler(func(r *http.Request) (interface{}, error) {
+		dr := dissectJSONRequest(r)
+		tlq, _ := helpers.ParseTimelineQuery(dr.FromQuery("q").AsString())
```
JoelAtDeluxe, line 158: A couple of points here:

1. You should handle the error.  This is user input, so it's reasonable to expect that the data would be unparsable. Just returning the error would be fine here.
2. Although `q` is pretty common as a name, we're currently using `query` for the same function in web. We should probably have consistent naming across the two rest endpoints

backend/server/api.go
```diff
@@ -152,6 +153,16 @@ func bindAPIRoutes(r *mux.Router, db *database.Connection, contentStore contents
...
+		i := services.ListEvidenceForOperationInput{
+			OperationSlug: dr.FromURL("operation_slug").Required().AsString(),
+			Filters:       tlq,
+		}
+		return services.ListEvidenceForOperation(r.Context(), db, i)
```
JoelAtDeluxe, line 163: The dissectJSONRequest function is really just a wrapper around accessing parts of the request. If those parts of the parsing don't work, then it'll _record_ an error. You still need to check if that error was produced. 

_Technically_, I think we can be pretty confident that `dr.Error` will be nil, as the URL parameter should be present, and the query parameter is optional. However, I'd still like to see the check here because by not including it we're setting ourselves up for an error down the road.

backend/server/api.go
JoelAtDeluxe:
```suggestion
		tlq, err := helpers.ParseTimelineQuery(dr.FromQuery("query").AsString())
		if err != nil {
			return nil, err
		}
```

backend/server/api.go
JoelAtDeluxe:
```suggestion
		if dr.Error != nil {
			return nil, dr.Error
		}
		return services.ListEvidenceForOperation(r.Context(), db, i)
```

JoelAtDeluxe (commented):
This makes sense to me. Is there any reason not to include searching for evidence here?

Another alternative: maybe we instead expose a detailed evidence view that builds on what `/api/operations/{op_slug}/evidence/{evidence_uuid}` does, but adds the missing info so we don't need to do a query over all evidence?

jrozner:
Originally the api endpoints were designed to be effectively write only as a way to allow submitting evidence from the public internet but only allowing viewing from a private network. This was due to some policies that were easier to adhere to that we no longer have to deal with internally. One of the things we've talked about is merging it all together like we have in the dev instance. I'd like to get some community feedback before merging this in to see if anyone else prefers that architecture. If not, we'll probably expand the api access to most/all endpoints that make sense. Is this a blocker for you right now?

audibleblink:
Not a blocker. We can maintain this API endpoint internally until a decision is made. 

The additional information was key to be able to keep evidence updates in our real-time chat, while keeping the report-bound content separate. 

I do see the appeal of the write-only API. Perhaps as an alternative, service workers could just be sent more data on evidence-creation.

jkennedyvz:
I'm asking for deployment requirements from our users to see if we can collapse the web/api containers for production deployments.

jrozner:
FYI, we're going to be merging the api and web servers (effectively the same as how the dev setup works) in the near future. We need to talk through exactly what that looks like but we expect that most endpoints will be available whether using session based authentication or api based authentication.

jkennedyvz:
This is closed with #898

PR #843, opened 2023-05-26
Tn/swap mux for chi

TylerNoblett:
After doing a bunch of research, it seemed that Chi was the best alternative to gorilla/mux, so I've swapped them out in this PR.


backend/authschemes/localauth/local_auth.go
```diff
@@ -77,13 +78,13 @@ func (LocalAuthScheme) Type() string {
...
 // the underlying system/database
-func (p LocalAuthScheme) BindRoutes(r *mux.Router, bridge authschemes.AShirtAuthBridge) {
+func (p LocalAuthScheme) BindRoutes(r chi.Router, bridge authschemes.AShirtAuthBridge) {
 	remux.Route(r, "POST", "/register", remux.JSONHandler(func(r *http.Request) (interface{}, error) {
 		if !p.RegistrationEnabled {
 			return nil, fmt.Errorf("registration is closed to users")
 		}
 
-		dr := remux.DissectJSONRequest(r)
+		dr := dissectors.DissectJSONRequest(r)
```
TylerNoblett, line 87: I'll put more info below, but the remux dissector functions aren't helpful anymore now that we're not using gorilla mux, so now the code just calls the dissectors directly

backend/server/dissectors/dissected_request.go
```diff
@@ -117,7 +116,11 @@ func (m *DissectedRequest) FromURL(key string) *Coercable {
 	if m.Error != nil {
 		return makeErrCoercable(m)
 	}
-	value, found := m.urlValues[key]
```
TylerNoblett, line 120: The biggest change is that gorilla Mux offers mux.Vars() (from which we created `urlValues`) which gives you all of the url keys and their values. Unfortunately, the closest `chi` comes is that you can use `chi.URLParam(m.Request, key)` to get a specific url param key. 

Because of this, I removed the urlParameters, as we can't use them anymore. I tried to honor the setup we had before, but manually creating that urlParam map with the chi functionality created a TON of boilerplate code for each request (it added two lines per each request, so hundreds of lines), so I just went with adding in this chi functionality here instead.

backend/server/dissectors/dissected_request_test.go
```diff
@@ -353,80 +353,23 @@ func TestParseBoolFlagFromQuery(t *testing.T) {
 }
 
 // Verify reading/converting from URL
-
-func TestParseStringFromURL(t *testing.T) {
```
TylerNoblett, line 357: Because we're not handling the urlparam map anymore, these tests all broke. I still thought it was important to test `URLParam`, which is basically now a wrapper for the chi function `chi.URLParam(m.Request, key)`, but there's some error handling and coercible stuff. However, I tried a bunch of methods and none of them worked (ie creating a mock implementation, monkey patching, copying the Chi's custom context and applying that in a test etc). The only way I could get it to work was passing in a second argument to the URLParam which was either chi or my fake implementation, but it seemed kind of over the top to add in a second arg to the 95 uses with chi, when this test is basically a wrapper for another library function, but also happy to do that if we think it would be beneficial!

backend/Readme.md
```diff
@@ -460,7 +460,7 @@ The details for this service are detailed in [pipeline readme](/backend/pipeline
 
 ## Development Overview
 
-This project utilizes Golang 1.13 (with modules), interfaces with a MySQL database and leverages Gorilla Mux to help with routing. The project is testable via docker/docker-compose and is also deployed via docker.
+This project utilizes Golang 1.13 (with modules), interfaces with a MySQL database and leverages Chi to help with routing. The project is testable via docker/docker-compose and is also deployed via docker.
```
JoelAtDeluxe, line 463:
```suggestion
This project utilizes Golang 1.20, interfaces with a MySQL database and leverages Chi to help with routing. The project is testable via docker/docker-compose and is also deployed via docker.
```

backend/server/api.go
```diff
@@ -9,40 +9,35 @@ import (
...
-	metricRouter.Handle("/api/metrics", promhttp.Handler())
-
-	r.Use(middleware.LogRequests(logger))
+func API(r chi.Router, db *database.Connection, contentStore contentstore.Store, logger logging.Logger) {
 	r.Use(middleware.AuthenticateAppAndInjectCtx(db))
+	r.Use(middleware.LogRequests(logger))
 
 	bindAPIRoutes(r, db, contentStore)
-	return r
+	r.Mount("/metrics", promhttp.Handler())
```
JoelAtDeluxe, line 28: I know we're talking about getting rid of metrics, so this may be moot, but does this use the logRequests and Authenticate middlewares? Ideally we'd want this endpoint to not have those.

backend/server/middleware/authenticator.go
```diff
@@ -83,7 +82,7 @@ func Policy(ctx context.Context) policy.Policy {
 	return p
 }
 
-func AuthenticateAppAndInjectCtx(db *database.Connection) mux.MiddlewareFunc {
+func AuthenticateAppAndInjectCtx(db *database.Connection) func(http.Handler) http.Handler {
```
JoelAtDeluxe, line 85: maybe it's worth defining a new type called MiddlewareFunc, which is the same as `mux.MiddlewareFunc` and `func(http.Handler) http.Handler`

 But if it's just these 3 uses, maybe it's not that valuable, as it's really just a way to keep the types simpler.

backend/server/remux/routing_wrappers.go
```diff
@@ -6,10 +6,10 @@ package remux
...
-	"github.com/gorilla/mux"
+	"github.com/go-chi/chi/v5"
 )
 
-// Route rewraps gorilla.mux's Handle/Methods to provide a better at-a-glance reading of route definitions
-func Route(r *mux.Router, method string, path string, handler http.Handler) {
-	r.Handle(path, handler).Methods(method)
+// Route rewraps chi's Handle/Methods to provide a better at-a-glance reading of route definitions
+func Route(r chi.Router, method string, path string, handler http.Handler) {
+	r.Method(method, path, handler)
```
JoelAtDeluxe, line 14: :neutral_face: This is basically what the `Route` function does. Probably best to do this the way you did it, but kind of annoying that we're just mirroring their approach.

backend/server/dissectors/dissected_request.go
JoelAtDeluxe: So, it looks like you  _can_ get all of the url parameters, but as a list of keys and a list of values (I don't know why it's not a list of keys _and_ values,  but I don't make the calls... at least here).

If we do want to go down that route, we'd call chi.RouteContext(r.Context()).URLParams

But, since we need this as a map, we either need to revise the dissectors to take the above type, or we need to create some code to re-wrap the values as `map[string]string`

backend/server/dissectors/dissected_request.go
```diff
@@ -117,7 +116,11 @@ func (m *DissectedRequest) FromURL(key string) *Coercable {
...
+	value := chi.URLParam(m.Request, key)
```
JoelAtDeluxe, line 119: Just spit balling here, but maybe instead of passing in a map, or using chi directly (thus tying this to a chi-only solution), we could instead take in a function to get a parameter, given a key?

backend/server/dissectors/dissected_request.go
TylerNoblett: ooh good find - yeah weird that it's two lists. Yeah I might just create some code to re-wrap the values, which would allow us to stick closer to the original code

backend/Readme.md
TylerNoblett: good catch!

backend/server/middleware/authenticator.go
TylerNoblett: if it was just two cases I don't think I would, but three is enough in my book to create a type

backend/server/remux/routing_wrappers.go
TylerNoblett: yeah I know right? I initially tried to follow their approach, but imo it made the code slightly less verbose, but it made finding a specific url path more difficult, so I shoehorned their approach into our code

backend/server/dissectors/dissected_request.go
TylerNoblett: oh yeah that would be better - I think I'm going to try to go back to the map of keys and values, which will change this code back to the way it was

backend/server/dissectors/dissected_request_test.go
TylerNoblett: added back in due to other changes

backend/authschemes/localauth/local_auth.go
TylerNoblett: no longer relevant

backend/server/middleware/authenticator.go
TylerNoblett: I added this to the bottom of authenticator.go - where would you put it? I googled around but didn't see any convention re where to put misc types

JoelAtDeluxe: Two perspectives:

* A lot of times I see types put at the top of a file where it gets used most. Or, if it'll have methods attached to it, then its own file. I think this works well some of the time, but there are times when you really just want to group data together, or create an alias (as we're doing here), and then i'm not sure it makes sense.
* Sometimes, for kind of miscellaneous types, I like to create a `types.go` file and put them all in there (in the same package), kind of mirroring what I've seen done in typescript. For smaller types _I_ think this helps, but to each their own.

I guess a 3rd perspective: You should do whatever you think is best. Putting this where it is now is probably fine, and can be moved later if someone has another idea about how it should be structured.

backend/server/api.go
TylerNoblett: fixed!

backend/server/middleware/authenticator.go
TylerNoblett: yeah I might go with the TS route!

TylerNoblett:
@JoelAtDeluxe I think I've addressed all of your feedback - thanks again for taking a look!

backend/bin/dev/dev.go
```diff
@@ -117,22 +118,22 @@ func tryRunServer(logger logging.Logger) error {
...
-	mux.Handle("/web/", http.StripPrefix("/web", server.Web(
-		db, contentStore, &server.WebConfig{
-			CSRFAuthKey:      []byte("DEVELOPMENT_CSRF_AUTH_KEY_SECRET"),
-			SessionStoreKey:  []byte("DEVELOPMENT_SESSION_STORE_KEY_SECRET"),
-			UseSecureCookies: false,
-			AuthSchemes:      schemes,
-			Logger:           logger,
-		},
-	)))
-	mux.Handle("/api/", server.API(
```
jkennedyvz, line 131: /web/ and /api/ should both be ported to the new router for dev.go

TylerNoblett: whoops! Just a second

jkennedyvz (approved):

PR #862, opened 2023-06-15
ensure only one key is deleted if multiple with same name

TylerNoblett:
Ensure that if there are multiple keys with the same name, only one is deleted

### DIFF
--- a/backend/helpers/functional.go
+++ b/backend/helpers/functional.go
@@ -16,7 +16,7 @@ func Map[T any, U any](slice []T, mapFn func(T) U) []U {
 	return result
 }
 
-// Find is a generic function that searches through a list searching for all items that match
+// Filter is a generic function that searches through a list searching for all items that match
 // the given predicate. This returns the elements that matched the predicate in the order they were
 // encountered.
 func Filter[T any](slice []T, predicate func(T) bool) []T {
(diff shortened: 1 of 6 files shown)


backend/helpers/functional.go
```diff
@@ -16,7 +16,7 @@ func Map[T any, U any](slice []T, mapFn func(T) U) []U {
 	return result
 }
 
-// Find is a generic function that searches through a list searching for all items that match
+// Filter is a generic function that searches through a list searching for all items that match
```
JoelAtDeluxe, line 19: Oops!

jrozner (changes requested):
It looks like this would remove the first key rather than the one selected. I think it would make sense to either add a unique constraint on the name (so this problem doesn't exist) or swap out the identifier in the DELETE request to be a unique identifier (like we do with slugs) so that we can target them directly.

TylerNoblett:
> It looks like this would remove the first key rather than the one selected. I think it would make sense to either add a unique constraint on the name (so this problem doesn't exist) or swap out the identifier in the DELETE request to be a unique identifier (like we do with slugs) so that we can target them directly.

oh that's a good point - yeah I feel like the first one makes the most sense. I'll take a look

TylerNoblett:
@jrozner we're now deleting based on the unique credential ID instead of the name

jkennedyvz (approved):

TylerNoblett:
@jrozner @jkennedyvz I just added a check to ensure that the username is unique

jrozner (approved):

PR #890, opened 2023-07-12
Tn/swap sessions for scp

TylerNoblett:
Switches out gorilla sessions for [scs](https://github.com/alexedwards/scs)


backend/authschemes/auth_bridge.go
```diff
@@ -128,8 +143,8 @@ func (ah AShirtAuthBridge) GetUserFromID(userID int64) (models.User, error) {
 
 // DeleteSession removes a user's session. Useful in situtations where authentication fails,
 // and we want to treat the user as not-logged-in
-func (ah AShirtAuthBridge) DeleteSession(w http.ResponseWriter, r *http.Request) error {
-	return ah.sessionStore.Delete(w, r)
+func (ah AShirtAuthBridge) DeleteSession(r *http.Request) {
```
TylerNoblett, line 146: `sessionManager.Remove` is a no-op if it can't find the session

backend/authschemes/auth_bridge_test.go
```diff
@@ -59,56 +60,50 @@ func TestCreateNewUser(t *testing.T) {
 
 type testSession struct{ Some string }
 
+func createContextForRequest(sessionManager scs.SessionManager, r *http.Request, w http.ResponseWriter) *http.Request {
```
TylerNoblett, line 63: necessary to mock a request sent through the SCS LoadAndSave middleware.

backend/authschemes/auth_bridge_test.go
```diff
@@ -59,56 +60,50 @@ func TestCreateNewUser(t *testing.T) {
...
 
 	gob.Register(&testSession{})
 
 	userID := createDummyUser(t, bridge, "")
 
 	browser := &testBrowser{}
 	w, r := browser.newRequest()
-	err := bridge.LoginUser(w, r, userID, &testSession{Some: "data"})
+	sr := createContextForRequest(*sessionManager, r, w)
+	err := bridge.LoginUser(w, sr, userID, &testSession{Some: "data"})
```
TylerNoblett, line 88: Note that in this file we're not actually seeing if the middleware works correctly (which is why we use the same request twice instead of creating a new request), but just ensuring that the bridge functions (`bridge.LoginUser`, `bridge.SetAuthSchemeSession` etc) don't error out, since they're basically wrappers for the SCS api

backend/authschemes/auth_bridge_test.go
```diff
@@ -59,56 +60,50 @@ func TestCreateNewUser(t *testing.T) {
...
-
-	browser := &testBrowser{}
-	w, r := browser.newRequest()
-	bridge.SetAuthSchemeSession(w, r, &testSession{Some: "data"})
-
-	_, r = browser.newRequest()
-	require.Equal(t, &testSession{Some: "data"}, bridge.ReadAuthSchemeSession(r))
-}
-
-func TestDeleteSession(t *testing.T) {
```
TylerNoblett, line 94: given that `bridge.DeleteSession(w, r)` is a wrapper for an scs function that doesn't return anything, there was really no reason to keep this test

backend/server/web.go
```diff
@@ -34,7 +36,7 @@ type WebConfig struct {
 	DBConnection     *database.Connection
 	AuthSchemes      []authschemes.AuthScheme
 	CSRFAuthKey      []byte
-	SessionStoreKey  []byte
+	SessionStoreKey  string
```
TylerNoblett, line 39: I repurposed the sessionstorekey to be the key name for the user sessions (the session is saved to each user's context, so I think naming it all the same thing shouldn't be a problem?), and the key must be a string

backend/services/user.go
```diff
@@ -480,26 +481,11 @@ func UpdateUserProfile(ctx context.Context, db *database.Connection, i UpdateUse
 	return nil
 }
 
-// DeleteSessionsForUserSlug finds all existing sessions for a given user, then removes them, effectively
-// logging the user out of the service.
-func DeleteSessionsForUserSlug(ctx context.Context, db *database.Connection, userSlug string) error {
```
TylerNoblett, line 485: This function was only getting called in a test, and isn't really usable with our new setup, so I deleted it

backend/session/store.go
```diff
@@ -4,79 +4,153 @@
...
+	fmt.Println("byteData: ", byteData)
+
+	return byteData, true, nil
 }
 
-func (store *Store) Set(w http.ResponseWriter, r *http.Request, s *Session) error {
-	sess := store.readRaw(r)
-	sess.Values[sessionDataKey] = s
-	return sess.Save(r, w)
+// Same function SCS uses to decode session data
```
TylerNoblett, line 82: didn't love that I had to copy a function from the scs library (since it could easily change), but due to their API I had to do this in order to save teh user ID into the database (since the user ID is how we are able to delete a session).

backend/authschemes/auth_bridge_test.go
```diff
@@ -59,58 +60,40 @@ func TestCreateNewUser(t *testing.T) {
...
 
-	_, r = browser.newRequest()
-	session := sessionStore.Read(r)
+	session := session.GetSession(sessionManager, sr)
 	require.NoError(t, err)
 	require.Equal(t, userID, session.UserID)
 	require.Equal(t, "data", session.AuthSchemeData.(*testSession).Some)
 }
 
-func TestAddToSession(t *testing.T) {
```
TylerNoblett, line 81: given that `bridge.SetAuthSchemeSession` is a wrapper for an scs function that doesn't return anything, there was really no reason to keep this test

TylerNoblett:
Closing Pull Request now that [Gorilla has new maintainers](https://github.com/orgs/gorilla/discussions/9)

PR #892, opened 2023-07-12
Create shared routes

TylerNoblett:
allow API access to most web routes. To do this, I put all the routes from web that either 1. weren't in API (excluding routes that use session data or authscheme data) or 2. were identical to the API into a shared routes file. Note that there were a few routes that had the same path in API and web but had slightly different implementations, so I left those in their respective route groupings.


jkennedyvz:
Hi @TylerNoblett, we want to add the API functionality into web, but not the other way around. 

We'll keep the web service as our only required container, and allow users that wish to keep the API service separate to do so. This should make transitioning deployments a bit easier as well.

backend/server/api.go
```diff
@@ -25,64 +24,17 @@ func API(r chi.Router, db *database.Connection, contentStore contentstore.Store,
...
 		bindAPIRoutes(r, db, contentStore)
 	})
 }
 
 func bindAPIRoutes(r chi.Router, db *database.Connection, contentStore contentstore.Store) {
-	route(r, "GET", "/operations", jsonHandler(func(r *http.Request) (interface{}, error) {
-		return services.ListOperations(r.Context(), db)
-	}))
-
 	route(r, "GET", "/checkconnection", jsonHandler(func(r *http.Request) (interface{}, error) {
```
jkennedyvz, line 34: We'll need /checkconnection available from web.go so clients can verify their configuration.

TylerNoblett:
> Hi @TylerNoblett, we want to add the API functionality into web, but not the other way around.
> 
> We'll keep the web service as our only required container, and allow users that wish to keep the API service separate to do so. This should make transitioning deployments a bit easier as well.

ah gotcha - I misunderstood. That makes sense!

TylerNoblett:
Resolved in #898

PR #897, opened 2023-07-26
Merge compose services

jkennedyvz:
Adding /api route to the dev frontend, prod frontend, and web binary to combine the web and api services. 

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #898, opened 2023-07-26
Tn/add api to web

TylerNoblett:
Based on John's feedback about not adding web functionality to the API, it was much easier for me to create a new branch than to correct my old [one](https://github.com/ashirt-ops/ashirt-server/pull/892). This branch ensures the shared routes only contains routes that are identical between API and web ~~(with the exception of `checkconnection`)~~ AND includes @jkennedyvz's config [changes](https://github.com/ashirt-ops/ashirt-server/pull/897). If this is what you guys prefer then we can close #892 and #897 

### DIFF
--- a/backend/server/shared_routes.go
+++ b/backend/server/shared_routes.go
@@ -1,0 +1,59 @@
+// Copyright 2023, Yahoo Inc.
+// Licensed under the terms of the MIT. See LICENSE file in project root for terms.
+
+package server
+
+import (
+	"net/http"
+
+	"github.com/ashirt-ops/ashirt-server/backend/contentstore"
+	"github.com/ashirt-ops/ashirt-server/backend/database"
+	"github.com/ashirt-ops/ashirt-server/backend/dtos"
+	"github.com/ashirt-ops/ashirt-server/backend/server/middleware"
+	"github.com/ashirt-ops/ashirt-server/backend/services"
+	"github.com/go-chi/chi/v5"
+)
+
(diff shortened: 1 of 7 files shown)


backend/server/shared_routes.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/ashirt-ops/ashirt-server/backend/contentstore"
+	"github.com/ashirt-ops/ashirt-server/backend/database"
+	"github.com/ashirt-ops/ashirt-server/backend/dtos"
+	"github.com/ashirt-ops/ashirt-server/backend/server/middleware"
+	"github.com/ashirt-ops/ashirt-server/backend/services"
+	"github.com/go-chi/chi/v5"
+)
+
+func bindSharedRoutes(r chi.Router, db *database.Connection, contentStore contentstore.Store) {
+	route(r, "GET", "/checkconnection", jsonHandler(func(r *http.Request) (interface{}, error) {
```
TylerNoblett, line 18: @jkennedyvz you [mentioned](https://github.com/ashirt-ops/ashirt-server/pull/892#discussion_r1274208048) adding checkconnection to web.go - I'm assuming we would still need this in the API routes as well (if someone chooses to keep them separate); if that's the case, then adding it to shared, should work, correct? 

Also, if this is added to web, how does that work? ie if you go to localhost:8080/checkconnection on the browser, there's no page so it just errors out. or would you be accessing it via web/checkconnection through something like postman?

PR #956, opened 2023-10-18
Send image urls

TylerNoblett:
Sends presigned url for images stored on s3, so that the backend doesn't have to load all of them before sending the evidence data


backend/contentstore/s3store.go
```diff
@@ -70,6 +71,23 @@ func (s *S3Store) Read(key string) (io.Reader, error) {
...
+func (s *S3Store) SendURL(key string) (*string, error) {
+	contentType := "image/jpeg"
+	req, _ := s.s3Client.GetObjectRequest(&s3.GetObjectInput{
+		Bucket:              aws.String(s.bucketName),
+		Key:                 aws.String(key),
+		ResponseContentType: aws.String(contentType),
+	})
+
+	// TODO TN how long should this be? and Should I try to cache these images
+	url, err := req.Presign(time.Hour * 24)
```
TylerNoblett, line 83: how long do you guys think these images should persist? I can't imagine someone leaving the application on the evidence page for more than an hour or so?

backend/contentstore/s3store.go
```diff
@@ -70,6 +71,22 @@ func (s *S3Store) Read(key string) (io.Reader, error) {
...
 
+func (s *S3Store) SendURL(key string) (*string, error) {
+	contentType := "image/jpeg"
+	req, _ := s.s3Client.GetObjectRequest(&s3.GetObjectInput{
+		Bucket:              aws.String(s.bucketName),
+		Key:                 aws.String(key),
+		ResponseContentType: aws.String(contentType),
+	})
+
+	url, err := req.Presign(time.Hour * 8)
```
TylerNoblett, line 82: do you guys have any thoughts about how long these images should persist?

TylerNoblett:
Alright! I'm able to the urls and pull them out of band, and nothing is broken on development, so I think it's good to go. Once you guys approve, I'll squash the commits down into one since it got out of hand

backend/contentstore/s3store.go
jrozner: maybe 5-10 minutes max. The browser is going to download these immediately and each request is going to create a new pre-signed url so there is no reason these need to be alive for very long. Two questions:
1. Can we limit the amount of times the evidence can be requested (1 or 2 is probably the right amount)
2. When you pop open the modal does it load a different version of the image or does it just render a scaled up version of the one that is already downloaded? If it's a new version are we requesting a new signed URL and downloading it?

jrozner:
Given that we support gpc as an alternative backend for storage, are we going to land the changes to support the equivalent there before merging?

TylerNoblett:
> Given that we support gpc as an alternative backend for storage, are we going to land the changes to support the equivalent there before merging?

Good question! I added the code for gpc this morning (in a separate branch), but haven't tested it yet. I would assume that these changes will be wayyy faster and worth it, but thought I would let you guys test the AWS side of things, before I bother Chris and get him to setup everything on gpc. But I can go ahead if you feel good about it

backend/contentstore/s3store.go
TylerNoblett: oh good point 👍 
1. As in to limit the number of times a user can reload the page in a certain time frame?
2. oooh good question - I hadn't thought about that. It is two requests with a new signed url. I can look into ways around that if you'd like

jkennedyvz: When you change your view in the timeline, like running a filter, you get new URLs to retrieve the images again. I think the most time someone would spend scrolling through a view would be around 30minutes or so? We could start there and increase if needed.

jkennedyvz: https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html indicates that pre-signed urls are intended to be multiple use, so limiting the time window they're active will be our best option.

jkennedyvz (approved):

PR #984, opened 2023-11-21
feat: ensure same url for full screen image

TylerNoblett:
Previously, after 30 minutes the initial presigned url for an image expired, and clicking on an image to view it full screen would generate a new presigned url every time (instead of creating a new presigned URL and using that for the duration of the url). This fixes that bug, and uses the same presigned url until it expires. I also updated the names of a few variables to match their functionality 

### DIFF
--- a/frontend/src/components/timeline/index.tsx
+++ b/frontend/src/components/timeline/index.tsx
@@ -89,6 +89,7 @@ export default (props: {
     <div className={cx('root')} ref={rootRef}>
       {props.evidence.map((evi, idx) => {
         const active = activeChildIndex === idx
+        console.log("evi active", evi.uuid, active)
         return (
           <TimelineRow
             {...props}
--- a/frontend/src/components/timeline/index.tsx
+++ b/frontend/src/components/timeline/index.tsx
@@ -83,6 +83,10 @@ export default (props: {
 
   const activeEvidence = props.evidence[activeChildIndex]
   if (activeEvidence == null) return null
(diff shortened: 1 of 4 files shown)


frontend/src/contexts/evidences_context/index.tsx
```diff
@@ -0,0 +1,41 @@
+import _ from 'lodash';
+import React, { PropsWithChildren, createContext, useContext } from 'react';
+import { UrlData } from 'src/global_types';
+
+interface IEvidencesContext {
+	imgDataSetter: (key: string, urlData: UrlData) => void
+	cachedUrls: Map<string, UrlData>
```
TylerNoblett, line 7: good use of Map!

frontend/src/components/timeline/index.tsx
```diff
@@ -36,7 +37,7 @@ export default (props: {
 
   const [activeChildIndex, setActiveChildIndex] = React.useState<number>(0)
   const [quicklookVisible, setQuicklookVisible] = React.useState<boolean>(false)
-  const [currImageUrlData, setCurrImageUrlData] = React.useState<UrlData| null>(null)
+  // const [currImageData, setCurrImageData] = React.useState<UrlData| null>(null)
```
TylerNoblett, line 40: Looks like this could be removed?

TylerNoblett (commented):
LGTM!

jkennedyvz (approved):

PR #1011, opened 2023-12-19
DRAFT: Enable remote debugging (DO NOT MERGE)

jrozner:
**DO NOT MERGE**

This enables remote debugging in the prod containers. We've noticed a recurring issue where the server becomes unresponsive after migrating from our EKS deployment to ECS after a period of time. We originally assumed this was related to a memory leak or file descriptor leak. After monitoring both it doesn't seem to be related. Let's get some container built that we can deploy and attach delve.

**DO NOT DEPLOY THESE UNLESS YOU KNOW WHAT YOU ARE DOING**

We might want to end up producing an image that is debugable and setup either dev or a separate docker compose file to use it.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,6 +44,7 @@ jobs:
     strategy:
       matrix:
         service: [frontend,web,api,init]
+        debug_mode: ["debug",""]
       fail-fast: false
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


jkennedyvz:
>We might want to end up producing an image that is debugable and setup either dev or a separate docker compose file to use it.

I added conditional debug builds for the web container, and tags for all containers in CI.  If you're fine with the new build config we can merge this.

PR #1041, opened 2024-02-16
Fix/key not set securecookie

luk3skyw4lker:
The bug was being caused by a change of validation in the securecookie package shown [here](https://github.com/gorilla/securecookie/compare/v1.1.1...v1.1.2#diff-107a4a38d8554908e24d329ea59692c4ca1d319af1d3048d1df8b614f47bede9L144)

The key being passed were 0 bytes in length, hence the error being thrown was: hash key not set. I have changed the tests to include a generic key and they're all working now, but a empty key is no longer accepted by securecookie.


jrozner (changes requested):
It looks like uuid got downgraded to 1.5.0. Can that get bumped back up before we merge this?

jrozner (approved):

PR #1042, opened 2024-02-16
fix: text editor stuck in 6 lines size

luk3skyw4lker:
Somehow a there was classless div between the evidence_preview codeblock and ace editor (code-viewer) divs that was limiting the height of the ace editor codeblock, all that was necessary was set this div height to 100% for everything to work.

The div between the other two mentioned divs:
image


jkennedyvz (approved):
LGTM!

PR #1044, opened 2024-02-22
feat: add adjusted at timestamp field to evidence

### LINKED ISSUE #894: Additional editable timestamp property on Evidence
Perfect evidence logging hygiene increases in difficulty as more people collaboratively work on an Operation.
Whether something was logged incorrectly, or just missed entirely, there's currently no first-class way to remedy this. As a workaround, the timestamp is just noted as a correction in the description field.

The preservation of the timeline is definitely a big value-add of ashirt's, so perhaps it's important to those who receive these timelines as a deliverable, to know that it hasn't been modified. So rather than editing an evidence's timestamp or inserting new evidence at arbitrary times, I propose an additional field in the Evidence model. 

Differentiation between when an action happened and when it was logged would give users the ability to view the timeline sorted by either.

Having a time-picker for a property of something like `occurred_at` would allow for fixups and late additions. Its default value in the modal could be the same as the `created_at` timestamp so as to not inconvenience users not needing this ability. The time-picker should exist in the web's evidence creation modal as well as the edit modal.

Too, it would be helpful if the operation's evidence page allowed for sorting of either timestamp. 
The new timestamp should also exist in exported information.

luk3skyw4lker:
Adding adjustedAt timestamp to evidence create and update process, which can be a nullable field, omitted on both operations.

Resolve #894


luk3skyw4lker:
Fixing broken tests already

frontend/src/components/date_time_picker/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+		setDate(new Date(newDate as Date))
+	}
+
+	const onChangeHour: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+		const [hours, minutes] = event.target.value.split(":").map(value => Number(value))
+		const newDate = date
+
+		newDate.setHours(hours, minutes)
+
+		setDate(new Date(newDate))
```
JoelAtDeluxe, line 38: I believe you could replace this with a datefns function:

```suggestion
		setDate(setHours(date, hours))
```

JoelAtDeluxe (commented):
Looks good to me. I spent more time on the Go code -- the only thing that seems to be missing would be updating the test seed to include some evidence with an adjusted date.

frontend/src/components/date_time_picker/index.tsx
luk3skyw4lker: I'll check it out, but probably I'll have to use setHours and setMinutes, whereas with the setHours function from the date object I can do it all in one call only

luk3skyw4lker:
@JoelAtDeluxe done and dusted, added adjustedAt for some evidences and all tests are running smoothly (at least locally)

JoelAtDeluxe:
@luk3skyw4lker looks good. Thanks a lot!

jkennedyvz (approved):

jkennedyvz (approved):

PR #1045, opened 2024-02-22
feat: prefix secret and access keys with AS-

### LINKED ISSUE #957: Prefix access keys and secret keys with a known value
Right now access keys and secret keys are just random values that are encoded. We should probably prefix the encoded value of each with a distinguishing prefix so that it's more easily possible to build secret detection rules for. We probably don't want the prefix to actually be used for the secret for signing to avoid any possible known prefix attacks.

luk3skyw4lker:
Prefixing keys with AS- after generation

Resolve #957

### DIFF
--- a/backend/server/api.go
+++ b/backend/server/api.go
@@ -21,7 +21,7 @@ func API(r chi.Router, db *database.Connection, contentStore contentstore.Store,
 	r.Handle("/metrics", promhttp.Handler())
 	r.Group(func(r chi.Router) {
 		r.Use(middleware.AuthenticateAppAndInjectCtx(db))
-		r.Use(middleware.LogRequests(logger))
+		// r.Use(middleware.LogRequests(logger))
 		bindSharedRoutes(r, db, contentStore)
 		bindAPIRoutes(r, db, contentStore)
 	})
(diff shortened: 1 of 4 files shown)


luk3skyw4lker:
reverting logger comments right now

jrozner (changes requested):
We need to add code here that will remove the `AS-` if and only if it exists from the secret key.

luk3skyw4lker:
@jrozner just to check if I understood: the prefix should only exists in the database then? Because then we would have to remove the "AS-" from every single usage of the key, is this right? That wasn't very clear at the PR, only that the prefix shouldn't be used for generation of the keys

jrozner:
The access key can have the `AS-` prefix, be stored in the database, and we don't need to do anything special. It's simply a value that we look up to retrieve the secret key associated with it. The secret key value is a key used for hmac. It needs to be a specific size and we don't want a known prefix for it. We can implement this in one of two ways:

1. Prepend `AS-` to the secret key and store it in the database then strip it off in the auth middleware before we pass it to the hmac function to compute the signature
2. Store the secret key in the database without the `AS-` prefix but display the secret key to the user on creation

In both cases, we'll need any client libraries to strip off the `AS-` before signing requests on the client side. I'm not sure what the best option is here and I think we run the risk of creating situations where bugs occur and it's not easy to debug. Maybe for now, let's just prepend the `AS-` to the access key but not the secret key and we can spend some time to decide what we want to do.

luk3skyw4lker:
@jrozner I agree, my concern was exactly the situation where it's hard to debug the problem, also, the secret key is stored in byte value so even if some key detection rule were added, ashirt would have to have two different matching patterns: a string and a byte one, and I think this isn't very practical for the secret key itself, especially since it's usage should only occur when authenticating the user.

jrozner (approved):

jrozner (approved):

PR #1047, opened 2024-02-22
feat: enforce dependencies for services in docker compose

### LINKED ISSUE #1012: Add dependencies and health checks to docker compose files
We're not currently doing health checks or enforcing dependencies and we should. I ran into an issue with #1011 where the prod compose file wouldn't start because the database wasn't up and running early enough.

luk3skyw4lker:
Adding dependencies and health checks for relevant services

Resolve #1012

### DIFF
--- a/docker-compose-prod.yml
+++ b/docker-compose-prod.yml
@@ -28,7 +28,7 @@ services:
 
   frontend:
     depends_on:
-      - backend
+      - ashirt-service
     build:
       context: .
       dockerfile: Dockerfile.prod.frontend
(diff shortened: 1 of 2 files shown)


jrozner (changes requested):
I think the nginx container needs to depend on the backend one because it checks the upstream exists at startup.

luk3skyw4lker:
@jrozner Isn't that the case already? All the frontend (nginx) containers depends on the backend to be up already

jrozner:
Ah sorry, I didn't realize we already had that.

jkennedyvz (approved):

jkennedyvz (approved):

jrozner (approved):

PR #1086, opened 2024-04-18
feat: c2-event evidence type

nbaertsch:
Adding an evidence type to capture C2 events in ASHIRT.

We have ingestors (python script + systemd service file) for both Cobalt Strike and Brute Ratel, though both have some bugs we are working through squashing. If we like the direction of adding this feature into ASHIRT I'd be partial to the ingestor repo being owned by ashirt-ops so that they can live with the rest of the project.

### DIFF
--- a/frontend/src/components/c2-event/index.tsx
+++ b/frontend/src/components/c2-event/index.tsx
@@ -179,7 +179,7 @@ export const C2EventEditor = (props: {
       </div>
       <div className={cx('extIP')}>
         <Input
-                label="Exeternal IP"
+                label="External IP"
                 className={cx('c2-event-input')}
                 value={props.value.externalIP || ''}
                 disabled={props.disabled}


frontend/src/global_types.ts
```diff
@@ -47,11 +48,32 @@ export type CodeBlock = {
...
+  c2: string,         // "Cobalt Strike"
+  c2Operator: string, // The c2 frameworke user that issued task
+  beacon: string,     // Beacon identifier
+  externalIP: string,
+  internalIP: string,
+  hostname: string,
+  userContext: string,// The user context that the implant/beacon/agent is running under
+  integrity: string,  // "Low", "Medium", "High", "System"
+  processName: string,// process image file shortname
+  processID: number,  // is 'number' acceptable here? Its a float :/
```
JoelAtDeluxe, line 62: Yeah, `number` could be fine here. That corresponds to a double sized float. 

But, are you sure this is a number at all? I don't know the format, but usually IDs are strings, even if they're only comprised of numbers. A number implies you can, say, add processIDs. Does it make sense to add these IDs?

nbaertsch: That being the case, string may be more appropriate.

nbaertsch:
> Saw this come in. Should have some time tomorrow to look through the PR. We're happy to adopt the ingestor repo over to the org. I think that makes sense as well. Do you have some samples of logs from brute ratel or cobalt strike so that we can test things end to end? Also, are the ingestors public currently?

Just made the ingestors public [here](https://github.com/humanaoffsec/ashirt-c2-ingestors). There are some (very succinct) sample logs there for BR and CS (no mythic atm), as well as a python script to simulate log writing to test the ingestors.

frontend/src/global_types.ts
nbaertsch: processID here is the PID of the beacon process - for context.

frontend/src/pages/operation_show/evidence_modals/index.tsx
```diff
@@ -145,53 +153,88 @@ export const EditEvidenceModal = (props: {
...
   const adjustedAtField = useFormField<Date | undefined>(props.evidence.adjustedAt ?? undefined)
-  React.useEffect(() => {
+  const c2EventField = useFormField<C2Event>({type: 'c2-event', c2: '', c2Operator: '', beacon: '', externalIP: '', internalIP: '', hostname: '', userContext: '', integrity: '', processName: '', processID: 0, command: '', result: ''})
+  if (props.evidence.contentType === 'codeblock') {
+    React.useEffect(() => {
+      getEvidenceAsCodeblock({
+        operationSlug: props.operationSlug,
+        evidenceUuid: props.evidence.uuid,
+      }).then(codeblockField.onChange)
+    }, [props.evidence.contentType, codeblockField.onChange, props.operationSlug, props.evidence.uuid])
```
JoelAtDeluxe, line 163: It's been a bit since I last did regular work in react, but I believe that we still want all of the hooks to always execute -- this shouldn't be in a conditional.

frontend/src/helpers/c2event_to_blob.ts
```diff
@@ -0,0 +1,38 @@
...
+    hostname: c2e.hostname,
+    userContext: c2e.userContext,
+    integrity: c2e.integrity,
+    processName: c2e.processName,
+    processID: c2e.processID,
+    command: c2e.command,
+    result: c2e.result,
+  }
+
+    evidence.metadata = {}
```
JoelAtDeluxe, line 35: Is there any reason you did this assignment outside of the main evidence creation above? Seems like you could have just done (omitting some lines here):

```typescript
  const evidence: JsonC2Evidence = {
    c2: c2e.c2,
    //...
    result: c2e.result,
    metadata: {}, // or c2e.metadata ?? {}
  }
```

frontend/src/components/c2-event/index.tsx
```diff
@@ -0,0 +1,259 @@
+// Copyright 2020, Verizon Media
+// Licensed under the terms of the MIT. See LICENSE file in project root for terms.
```
JoelAtDeluxe, line 2: I think the majority of copyright headers have been removed, and we could probably drop the header. @jrozner  or @jkennedyvz can verify.

frontend/src/components/c2-event/index.tsx
```diff
@@ -0,0 +1,259 @@
...
+import { C2Event } from 'src/global_types'
+import ComboBox from 'src/components/combobox'
+import { useAsyncComponent } from 'src/helpers'
+import LoadingSpinner from 'src/components/loading_spinner'
+import WithLabel from 'src/components/with_label'
+import classnames from 'classnames/bind'
+const cx = classnames.bind(require('./stylesheet'))
+
+
+export const C2EventTextArea = React.forwardRef((props: SharedProps & {
```
JoelAtDeluxe, line 15: Why did you decide to create a text area here, rather than reuse the one from `components/input` ?

frontend/src/components/c2-event/index.tsx
jkennedyvz: That's correct @JoelAtDeluxe we no longer use copyright headers in each file.

jrozner (changes requested):
Overall things seem to work. I wasn't able to actually test with real data using the ingestors because the `ashirt_worker` import is failing. Is that a separate file that you have created that isn't included within the repo?

frontend/src/pages/operation_show/evidence_modals/index.tsx
nbaertsch: I had this quite messed up in the original commit. I've fixed the useEffect issue and cleaned up the return logic.

frontend/src/helpers/c2event_to_blob.ts
nbaertsch: This is fixed.

frontend/src/components/c2-event/index.tsx
nbaertsch: Thats a very good question. I missed that component entirely when I was writing this initially. I've gone ahead and removed the custom TextArea.

nbaertsch:
I've made the changes we've discussed here. Sorry for the delay on this one, thanks for your patience. If there's anything else I can do to clean this up do lmk, this was my first time writing React 😁

jrozner:
Thanks for getting the changes in. I should have some time this afternoon to take a look

frontend/src/components/input/index.tsx
```diff
@@ -4,7 +4,7 @@ import WithLabel from 'src/components/with_label'
 import classnames from 'classnames/bind'
 const cx = classnames.bind(require('./stylesheet'))
 
-type SharedProps = {
+export type SharedProps = {
```
jrozner, line 7: Does this need to be exported?

frontend/src/pages/operation_show/evidence_modals/index.tsx
```diff
@@ -145,18 +153,28 @@ export const EditEvidenceModal = (props: {
   const tagsField = useFormField<Array<Tag>>(props.evidence.tags)
   const codeblockField = useFormField<CodeBlock>({ type: 'codeblock', language: '', code: '', source: null })
   const adjustedAtField = useFormField<Date | undefined>(props.evidence.adjustedAt ?? undefined)
+  const c2EventField = useFormField<C2Event>({type: 'c2-event', c2: '', c2Operator: '', beacon: '', externalIP: '', internalIP: '', hostname: '', userContext: '', integrity: '', processName: '', processID: '', command: '', result: ''})
   React.useEffect(() => {
-    if (props.evidence.contentType !== 'codeblock') {
+      if (props.evidence.contentType !== 'codeblock' && props.evidence.contentType !== 'c2-event') {
```
jrozner, line 158: Would you mind running this file through a formatter? It looks like there is an extra indent here and some formatting changes below

jrozner (commented):
Overall things look pretty good. @JoelAtDeluxe any chance you have time to look at the react components? I don't have much experience with it and can't provide meaningful feedback for idiomatic aspects.

JoelAtDeluxe:
I'll make some time this weekend to go through it.

JoelAtDeluxe (commented):
Still somewhat going through this, but there are enough things here that I'd like to see changed. I can try to find time to help if you need it.

Some big thoughts:
1. There are a few styling issues to resolve: the grid is too big (when viewing the data, the right side goes off screen), the box shadows aren't rendering correctly (only the top renders), when editing, the grid is off center.
2. The editor and viewer are nearly identical. Maybe it makes sense to combine them into one (unexported), and create a simple component for reader and writer?
3. As Joe noted, there are some files with formatting issues. I agree that this should probably be run through a formatter. I usually use VSCode to do this for me at the file level. There are also some other issues around extraneous imports. You could also run eslint to help identify a few of these issues (in frontend, run `npm run lint`), but it looks like there errors in other files outside of this PR too. :disappointed: 

It looks like basically all of the logic is correct at this point though, and I didn't see any issues specifically around React, so it seems like this PR is really just in a cleanup phase.

nbaertsch:
Thanks for the review Joel. I'm going to get to work on the code cleanup work you noted. When it comes to styling I'd appreciate some help (especially grid sizing issue, i can't get that thing centered for the life of me), but let me give this one more pass and I'll ping for another review and any styling help I may need.

JoelAtDeluxe:
I'll try to find some time. fit seems like the primary issue why it doesn't fit right is because of the padding on the c2-event-grid. That somewhat breaks the lightbox view, but I think that there should be a way to fix specifically that (it's been awhile though, so I don't know what it is off the top of my head).
I'd say look at the har viewer as an example, but that's broken due to lazy loading components. There may still be some insight there though.

PR #1194, opened 2024-10-07
Remove go-kit and switch to slog

### LINKED ISSUE #1191: Logging overhaul part 1
We currently bring in go-kit for structured logging and have a bunch of helpers to support using it due to the lack of structured logging when we created this. The `log/slog` package now exists and we don't really need to bring in this external dependency anymore. Let's replace it with the std `slog` package.

For ease we'll do this in two parts because of differences between the `go-kit/logging` package from `log/slog`. In the first part we'll rip out `go-kit/logging` and replace all the instances with `log/slog` but keep all the existing helpers, replacing the underlying type.

jrozner:
This removes go-kit's logging and switches to slog.

Closes #1191

### DIFF
--- a/backend/logging/logging.go
+++ b/backend/logging/logging.go
@@ -2,52 +2,44 @@ package logging
 
 import (
 	"context"
+	"io"
+	"log/slog"
 	"os"
 
-	kitlog "github.com/go-kit/kit/log"
 	"github.com/google/uuid"
 )
 
-// Logger is a generic logging interface. Currently wraps Go-kit's log.Logger
-type Logger kitlog.Logger
-
 var requestLoggerKey = &struct{ name string }{"requestLogger"}
 
-var systemLogger kitlog.Logger
+var systemLogger *slog.Logger
 
 // SetupStdoutLogging creates a new logger that logs to standard out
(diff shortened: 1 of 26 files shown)


backend/logging/logging.go
```diff
@@ -2,52 +2,44 @@ package logging
...
+	"io"
+	"log/slog"
 	"os"
 
-	kitlog "github.com/go-kit/kit/log"
 	"github.com/google/uuid"
 )
 
-// Logger is a generic logging interface. Currently wraps Go-kit's log.Logger
-type Logger kitlog.Logger
```
JoelAtDeluxe, line 12: I probably would have kept Logger and just had it wrap *slog.Logger to minimize the differences, but this works too.

JoelAtDeluxe (approved):
Yeah, looks good.

backend/logging/logging.go
jrozner: I originally tried that and ran into some type issues. This ended up being the easiest way to solve it. My hope is that we won't ever need to swap this again so it won't matter but I agree, that would have made it probably simpler.

PR #1277, opened 2025-07-23
Updates without csrf

jrozner:


jrozner:
This is currently broken due to https://github.com/stylus/stylus/issues/2938. main is broken too. We need to wait for this to be resolved or remove stylus from the frontend dependencies.

PR #1462, opened 2026-06-17
Remove AWS Lambda enhancement-service support

jrozner:
## Summary

The enhancement-services subsystem supported two interchangeable worker transports selected by a `type` field in each worker's JSON config: `web` (HTTP) and `aws` (AWS Lambda, plus a local RIE dev path). The `aws` type was never surfaced in the admin UI (which is a raw-JSON editor) and is functionally redundant — a Lambda can be reached as a `web` worker over an HTTP Function URL or API Gateway.

This PR drops the Lambda transport and standardizes on a single HTTP transport, then simplifies the now-single-implementation code.

## Changes

- **Delete** `worker_aws.go`, `rie_interface.go`, and their tests (Lambda worker + RIE local emulator).
- **Drop** the `github.com/aws/aws-sdk-go-v2/service/lambda` dependency. The S3 SDK usage in `internal/contentstore` is unaffected and remains.
- **Collapse the abstraction**: retire the `ServiceWorker` interface and the `findAppropriateWorker(...) → handler.Build(...)` two-step in favor of a single `buildWorker()` constructor. It still validates `type`/`version`, so any legacy `aws` config now fails with a clear `unsupported service worker type "aws" (version 1)` error instead of the old generic message.
- **Remove** the `UseLambdaRIE` config field/getter and the `APP_USE_LAMBDA_RIE` docker-compose var.
- **Simplify tests**: relocate the shared mock helpers out of the deleted RIE test file, drop the `transformFn`/`wrapInAwsResponse` Lambda-envelope wrapping (only HTTP remains), rename `awsProcessResp` → `processResp`.
- **Docs**: update `pipeline_readme.md` to note that HTTP-exposed Lambdas are still usable as `web` workers; remove the `lambda/js-container` template link.

Net ~580 lines removed.

## Compatibility

The worker config schema (`{type, version, url, headers}`) is **unchanged**, so **no DB migration is required**. Async/deferred processing remains available over HTTP via the existing 202-Accepted / `"deferred"` response handling.

### Operator action required

Any deployment with workers configured for the Lambda transport must repoint them at an HTTP endpoint (e.g. a Lambda Function URL). Audit existing workers with:

```sql
SELECT id, name, config FROM service_workers
WHERE deleted_at IS NULL AND config LIKE '%"aws"%';
```

## Verification

- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./internal/enhancementservices/...` — pass (web Test/ProcessMetadata: success / error / deferral / rejection)


jrozner:
@dependabot rebase

